Issue #75, opened 2019-06-08
Different image sorting than by name


karlch:
Currently images are always sorted by name / path. There could be different sorting, e.g. case-sensitive vs. case-insensitive, by time (exif, access, ...) and so forth.

Implementation could be a new setting `image.order` or similar, which has different options. The options would be converted to a function that takes the image path an returns the sort key so it can be passed to `sorted`.

Therefore there would be two main steps:
* Write a new `ImageOrder` setting
* Use the setting value in the calls to `sorted`, one in `utils.files.listdir` for library, opening in the beginning, one in `api.commands._CommandArguments.parse_args` for path-like commands

Thanks for the initial idea in https://github.com/karlch/vimiv/issues/79

Issue #30, opened 2019-01-27
Improve logging


karlch:
The current usage of logging is a bit inconsistent and could be improved by:
- [x] Using the log levels consistently according to [the python docs](https://docs.python.org/3.7/howto/logging.html#logging-basic-tutorial)
- [ ] Adding further debug statements
- [x] Refining the statusbar log handler, it is currently not very useful, as the messages are often removed very quickly

karlch:
This was partly addressed in #105 but quite a few modules still miss a significant amount of debug statements. Will probably be done gradually when debugging existing issues.

karlch:
Closing as this has become stale and will almost certainly only be worked on "on-the-fly", not explicitly to solve this issue.

PR #76, opened 2019-06-08
WIP: [ImageOrder]: #75 added ImageOrder setting

kaldown:


kaldown:
Actually I don't understand the commentary about `api.commands._CommandArguments.parse_args`, this call won't use `listdir`, so I placed it into `_get_content`.
What do you think, any suggestions? ;)

kaldown:
I just don't get how to change state of a settings variable and then call `_reload_dir()` =(
Have you any examples?

karlch:
Wow, thanks for working on this so quickly :tada: 

I quickly looked at the code and have a few hints if you don't mind :)

With your implementation images are now always sorted by mtime, I would actually prefer the alphabetical order to stay the default and the mtime to be an option. Instead of storing a boolean, the new ImageOrder setting you introduced could have a callable function as value that is used as sort key depending on the name. This could for example be implemented by a dictionary mapping names to callables and return them accordingly (mapping name to a function that returns the name and mtime to `os.path.getmtime` as you currently use it).

If you would like to add reverse to this, this could either be an additional boolean setting or somehow mingled with the `ImageOrder` setting.

Concerning reloading: All settings have a `changed` signal. The working directory handler could connect a method to this signal and reload the current directory by calling its own `_reload_directory` method in the connected slot. There are quite a few examples of slots connected to changed settings. One would be in `gui.library` where the library connects to the `library.width.changed` signal and calls `_on_width_changed` to update accordingly.

Hope this helps and thanks again for your work! :)

kaldown:
now it should work, but still can't find the problem with
` INFO: No option 'image_order' in section: 'GENERAL' in configfile`
I've tried to mimic nearly constructed `monitor_fs` and don't see any difference, but somehow this warn appears.

karlch:
Thanks again, seems to work well now :)

The info is just appearing because you probably haven't added the setting to your local `vimiv.conf` file. Initial idea was to make new settings quickly visible, but I may have to re-think this as it has lead to confusion before. This is part of #30.

Three minor points:
* You have introduced a new `:order` command in the library, but in principle there is already the `:set` command for changing the value of settings.
* Going with this, setting it with `:set` doesn't work with your reversing syntax as the `-` is treated specially by argparse.
* You may want to rename the setting to `file_order` or similar as it now also applies to the directories in the library (which is better than my initial idea!)

Maybe you could update this? Otherwise looking very good!

kaldown:
Yup, of course I will re-do this PR.
Thanks for a feedback

karlch:
Perfect, thanks! If you like you can also add yourself to the `AUTHORS` file :)

karlch:
Closing after #563

Issue #100, opened 2019-08-17
Extend information displayed by :metadata


karlch:
Currently the exif information shown by the `:metadata` command is rather limited. Mainly because I am not very experienced with exif and don't know what information is useful to camera-lovers and how it should be formatted. Therefore this is a great place for anyone with more knowledge to step in.

The information is retrieved in the `ExifInformation` dictionary class in `imutils.exif`.  It contains a dictionary with all exif information retrieved by [piexif](https://piexif.readthedocs.io/en/latest/). The `ExifInformation` itself is filled with keys corresponding to the displayed name of the information and their respective values. Example:
```
{"Make": "Canon"}
```

All of the information loading happens in the `_load_exif` method which gets called upon construction. The `MetadataWidget` in `gui.metadata_widget` knows how to process and display the information in this dictionary.

To extend the information displayed only more information must be added to the dictionary, making sure to catch any `KeyError` exceptions in case the exif data is not available for the current image.

eorus:
I would like to ask if there is any update on this one. I want to see the image size info or a meta deta display. Is there a way to get this information in the current build?

karlch:
Thanks for your interest. There are currently two options:
* The `:metadata` command, also bound to `i` in image mode which shows some basic EXIF information. Adding the image size here should not be difficult if this is wished.
* The `image-size` status module which would display the image size in the statusbar. To use it, add `{image-size}` anywhere in one of the image mode status configurations in the `vimiv.conf`, e.g.:
```
[STATUSBAR]
...
left_image = {index}/{total} {basename} [{zoomlevel}] {image-size}
```

If you are interested in any specific other metadata, just let me know. If you like, you could also extend the code `imutils.exif`, especially the loading of the dictionary around `ExifInformation._load_exif`.

Note that the `:metadata` command requires [piexif](https://github.com/hMatoba/Piexif) to be installed as it used this module to load the data.

Hope this helps!

eorus:
Adding {image-size} simply solves this issue. Metadata can be used for more information. Thanks for the quick support and for Vimiv. 

This is pure speed and navigation image viewer, probably the fastest I have used so far for browsing images.

karlch:
Thanks for these warm words, just let me know if you need help with anything else or have any further suggestions.

karlch:
As the information is now fully configurable thanks to #212, this is much closer to a solution.

[...]

PR #212, opened 2020-05-18
Metadata Handling

jeanggi90:
Related to #100. I do not think that there is one set of metadata which suit the need to all users. Neither do I think that all users want to fiddle with the source code in order to get their desired metadata to show. Therefore I have added an option to set the desired metadata as a list in the settings.
Since the data is treated generically, I have not yet figured out the best was to properly format the data, e.g. add units, as `piexif` does not seem to have a notion of the units themselves.

Furthermore, depending on the stage of my photography workflow, different metadata information are relevant. Therefore I have thought about adding 3 to 5 user customizable sets of metadata key, which can be viewed one the metadata widget. Changing the metadata set should be done using a shortcut (e.g. <kbd>CTRL</kbd> + <kbd>1</kbd> through <kbd>CTRL</kbd> + <kbd>5</kbd>).

@karlch tell me what you think about that

### DIFF
--- a/vimiv/api/settings.py
+++ b/vimiv/api/settings.py
@@ -488,3 +488,14 @@ class title:  # pylint: disable=invalid-name
         desc="Default window title if no mode specific options exist",
     )
     StrSetting("title.image", "vimiv - {basename}", desc="Window title in image mode")
+
+
+class metadata:  # pylint: disable=invalid-name
+    """Namespace for metadata related settings."""
+
+    # Store the keys as a comma seperated string
+    StrSetting(
+        "metadata.keylist",
+        "Make, Model, DateTime, ExposureTime, FNumber, isospeedratings, FocalLength",
+        desc="Define the metadata keys to display"
+    )
--- a/vimiv/api/settings.py
+++ b/vimiv/api/settings.py
@@ -497,5 +497,5 @@ class metadata:  # pylint: disable=invalid-name
     StrSetting(
         "metadata.keylist",
         "Make, Model, DateTime, ExposureTime, FNumber, isospeedratings, FocalLength",
-        desc="Define the metadata keys to display"
+        desc="Define the metadata keys to display",
     )
--- a/vimiv/api/settings.py
+++ b/vimiv/api/settings.py
@@ -494,8 +494,8 @@ class metadata:  # pylint: disable=invalid-name
     """Namespace for metadata related settings."""
 
     # Store the keys as a comma seperated string
-    StrSetting(
-        "metadata.keylist",
+    keys = StrSetting(
+        "metadata.keys",
         "Make, Model, DateTime, ExposureTime, FNumber, isospeedratings, FocalLength",
         desc="Define the metadata keys to display",
     )
--- a/vimiv/imutils/exif.py
+++ b/vimiv/imutils/exif.py
@@ -101,38 +101,28 @@ class ExifInformation(dict):
             self._exif = piexif.load(filename)
         except (piexif.InvalidImageDataError, FileNotFoundError, KeyError):
             return
-        # Read information from 0th IFD
-        ifd = piexif.ImageIFD
-        self._add_bytes_info("Make", "0th", ifd.Make)
-        self._add_bytes_info("Model", "0th", ifd.Model)
-        self._add_bytes_info("DateTime", "0th", ifd.DateTime)
-        # Read information from Exif IFD
-        ifd = piexif.ExifIFD
-        self._add_two_digits("ExposureTime", "Exif", ifd.ExposureTime, suffix=" s")
-        self._add_fraction("FNumber", "Exif", ifd.FNumber, prefix="f/")
-        self._add_fraction("ShutterSpeedValue", "Exif", ifd.ShutterSpeedValue, " EV")
-        self._add_fraction("ApertureValue", "Exif", ifd.ApertureValue, " EV")
-        self._add_fraction("ExposureBiasValue", "Exif", ifd.ExposureBiasValue, " EV")
-        self._add_fraction("FocalLength", "Exif", ifd.FocalLength, " mm")
-        _logger.debug("%s initialized, content:\n%r", self.__class__.__qualname__, self)
 
-    def _add_bytes_info(self, name, ifd, key):
-        """Add exif information of type bytes to the dictionary."""
-        with contextlib.suppress(KeyError):
-            self[name] = self._exif[ifd][key].decode()
+        for ifd in self._exif:
+            if ifd == "thumbnail":
+                continue
 
-    def _add_fraction(self, name, ifd, key, suffix="", prefix=""):
-        """Add exif information as fraction of two numbers to the dictionary."""
-        with contextlib.suppress(KeyError):
-            value_tuple = self._exif[ifd][key]
-            fraction = value_tuple[0] / value_tuple[1]
-            self[name] = f"{prefix}{fraction:.2f}{suffix}"
-
-    def _add_two_digits(self, name, ifd, key, separator="/", suffix="", prefix=""):
-        """Add exif information as two numbers to the dictionary."""
-        with contextlib.suppress(KeyError):
-            value_tuple = self._exif[ifd][key]
-            self[name] = f"{prefix}%s{separator}%s{suffix}" % value_tuple
+            for tag in self._exif[ifd]:
+                keyname = piexif.TAGS[ifd][tag]["name"]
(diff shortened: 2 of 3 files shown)


karlch:
Thanks for this PR, I like it a lot!

I am also not quite sure what to do concerning the units. To be fair I am also fine with just leaving them out. If you think this is not an option, we could keep our own dictionary mapping exif key names to their corresponding units, falling back to no unit. So something along:
```
units = {
    "ExposureTime": "s",
}
```
Or even with the corresponding structure from piexif:
```
units = {
    piexif.ExifIFD: {
        piexif.ExifIFD.ExposureTime: "s",
    }
}
```
and them add the units on-demand / when we notice something is missing. I am not sure the maintenance overhead is worth it though. As much as I should run aways from numbers without units as a physicist :smile: 

Now to your second point, the user-configurable sets of metadata keys. I also find this a brilliant idea, but would recommend to put this into a second PR after the current one to keep it simple as this is probably quite a bit more effort. If you don't mind, I would open a new issue for this where I could give some basic hints on a possible implementation. You could then take over if you like.

If it is not too much to ask, I would leave a few comments concerning the PR code that could be addressed before merging. It would be great if you could also fix the issues raised by the `lint` ci. Most of them should be gone once [running pre-commit over the code](https://karlch.github.io/vimiv-qt/documentation/hacking#style-and-formatting). Don't worry if you don't get to it, I can also fix this after merging.

Thanks again, much appreciated!

jeanggi90:
I am glad that you like my idea as well as implementation. I must say it is a real pleasure to work with your clean, well structured and well documented code!

The unit is actually not the only problem, but the formatting of the data is also a not right. For example the shutterspeed is normally given as a fraction like `1/250 s` whereas the f-stop is given in floats like `f/5.6`. In my view, this formatting (including) the units is something which should be provided by the metadata reader tool. For example the [exiv2](https://www.exiv2.org/) does a good job at that. However, in the [python wrapper](https://python3-exiv2.readthedocs.io/en/latest/index.html) I did not find this option. 
Anyways, if you are fine with how it is at the moment, I would suggest that we leave it as it is, but open and issue which addresses the formatting and the units of the metadata.

I think it makes sense to open a second PR for my other idea. Especially, since I am not reluctant in getting additional information and hints concerning the implementation (esp. how to deal with the different information displayed in the metadata widget and how to add new commands/keybindings etc.) :blush: 

Back to the current PR. If you agree, I will do the final cleanup, fix the linting error and then remove the draft flag.

vimiv/api/settings.py
```diff
@@ -488,3 +488,14 @@ class title:  # pylint: disable=invalid-name
         desc="Default window title if no mode specific options exist",
     )
     StrSetting("title.image", "vimiv - {basename}", desc="Window title in image mode")
+
+
+class metadata:  # pylint: disable=invalid-name
+    """Namespace for metadata related settings."""
+
+    # Store the keys as a comma seperated string
+    StrSetting(
```
karlch, line 497: It would be great to asign the setting to a variable, e.g.: `keylist = StrSetting( ...`. Like this we can refer to it later via the python object, not only as a string.

vimiv/imutils/exif.py
```diff
@@ -99,40 +100,42 @@ def _load_exif(self, filename):
...
+            for tag in self._exif[ifd]:
+                keyname = piexif.TAGS[ifd][tag]["name"]
+                keytype = piexif.TAGS[ifd][tag]["type"]
+                val = self._exif[ifd][tag]
+                _logger.debug(
+                    f"name: {keyname} type: {keytype} value: {val} tag: {tag}"
+                )
+                if keyname.lower() not in desired_keys:
+                    _logger.debug(
+                        f"{keyname.lower()} not in {desired_keys}. Ignoring it"
```
karlch, line 124: Maybe another reference to `{desired_keys}` is a bit too verbose. I would personally prefer to just have "desired keys" as a string, but leaving it as it is is fine too.

vimiv/imutils/exif.py
```diff
@@ -99,40 +100,42 @@ def _load_exif(self, filename):
...
+                    )
+                    continue
+                if keytype in (1, 3, 4, 9):  # integer
+                    with contextlib.suppress(KeyError):
```
karlch, line 128: As far as I can see all of these calls to `contextlib.suppress` are not needed as we are just adding a value to the dict, not retrieving anything. Thus `KeyError` should never be raised, or am I missing something?

vimiv/imutils/exif.py
```diff
@@ -99,40 +100,42 @@ def _load_exif(self, filename):
...
+                val = self._exif[ifd][tag]
+                _logger.debug(
+                    f"name: {keyname} type: {keytype} value: {val} tag: {tag}"
+                )
+                if keyname.lower() not in desired_keys:
+                    _logger.debug(
+                        f"{keyname.lower()} not in {desired_keys}. Ignoring it"
+                    )
+                    continue
+                if keytype in (1, 3, 4, 9):  # integer
```
karlch, line 127: Concerning the keytype, are you sure concerning these numbers? Would it be possible to use the values from `piexif.TYPES` instead, so we don't rely on the integer values? So something like:
```
if keytype in (piexif.TYPES.Short, piexif.TYPES.Long, piexif.TYPES.SShort, piexif.TYPES.SLong):
    ....
```

vimiv/imutils/exif.py
```diff
@@ -99,40 +100,42 @@ def _load_exif(self, filename):
...
+                        self[keyname] = val
+                elif keytype == 2:  # byte encoded ascii
+                    with contextlib.suppress(KeyError):
+                        self[keyname] = val.decode()
+                elif keytype in (5, 10):  # (int, int) <=> numerator, denominator
+                    with contextlib.suppress(KeyError):
+                        self[keyname] = f"{val[0]}/{val[1]}"
+                elif keytype == 7:  # byte
+                    with contextlib.suppress(KeyError):
+                        self[keyname] = val.decode()
```
karlch, line 138: This is the same logic as above for the byte encoded ascii, so maybe they could be merged with a `keytype in` like for the other values?

vimiv/imutils/exif.py
```diff
@@ -99,40 +100,42 @@ def _load_exif(self, filename):
         """Load exif information from filename into the dictionary."""
         try:
             self._exif = piexif.load(filename)
+            desired_keys = [
+                e.lower().strip()
+                for e in api.settings.get_value("metadata.keylist").split(",")
```
karlch, line 105: If the setting was stored in a variable, you could refer to it here via the object. I.e.:
```
for e in api.settings.metadata.keylist.value.split(", ")
```
I admit this is quite a list of `.` chained stuff, but at least we don't rely on the string.

vimiv/api/settings.py
jeanggi90: Surely, I have noticed that you sometimes assign the setting to a variable and sometimes you dont. May I ask you want the advantage or point of doing in one or the other way?

vimiv/imutils/exif.py
jeanggi90: You are right. I did not think about that. Thanks for pointing that out!

vimiv/imutils/exif.py
jeanggi90: I see, where it comes handy to store the setting in a variable. :blush:  However, I would suggest to add an`.strip()`.This way it is more robust when do user does not get the formatting fully right

vimiv/api/settings.py
karlch: The general rule is to use the variable where possible as we can then make use of actual python objects directly. This also gives us access to the `updated` signal which allows objects to react when a setting changed, e.g. the slideshow in `imutils.slideshow` has a timer that updates its interval upon `api.settings.slideshow.delay.changed`.

Of course there are places where we cannot use the type immediately, for example when running the `:set` command on the command-line, we first retrieve the object by its name as only the name can be passed as a string by the user.

If there are only cases where the setting is used by name and it cannot be used by its type, the setting is not assigned to al variable. This is currently only the case for status-related settings, for example `statusbar.left` and `statusbar.left_image`. Here we can define arbitrary combinations of `position_mode`, but not all have defaults. We also only retrieve the value via this string pattern as we can then iterate over the positions and make use of the current mode name. This happens for example in `gui.statusbar` in the `_get_text(position, mode)` method.

tl;dr assign where possible, use string where the code is too dynamic to access the variable directly

Hope this makes it a bit more clear instead of more confusing :smile:

karlch:
Thanks again! :tada:

Issue #213, opened 2020-05-19
Add customizable sets of metadata information


karlch:
This could be a follow-up to #212 proposed by @jeanggi90.

The idea is to have a customizable set of metadata information which can be switched easily at runtime. Initial idea was to use `<ctrl>1` to `<ctrl>5`. Personally I would prefer to use `count` directly with the `:metadata` binding, i.e. `1i` to `5i` would toggle between the options, `i` without a `count` would toggle the actual widget. This would be consistent with the slideshow command. What do you think @jeanggi90?

Some hints concerning the implementation:

First we would need the options somewhere, e.g. in the new `metadata` settings class:
```
options = {
    1: "first set of options",
    2: "second set of options"
}
```

Then we would need to populate these with values from the config file in case they are defined. I thought of a new function in `config.configfile` that would somehow iterate over the `METADATA` section, similar to the `_read_plugins` function. How exactly these options are defined is up to discussion. Something simple could be:
```
[METADATA]
0 = first set of options
1 = second set of options
```
which could be parsed quickly by checking if the config key can be converted to an integer.

Finally, the metadata widget in `gui.metadata` would need to know about all this. First of all the `metadata` function which serves as the command would need the `count` argument. Then in the body, a check for `if count is not None` followed by the actual config switching. Something along `api.settings.metadata.keylist.value = api.settings.metadata.options[count]` with proper error checking.

To actually get the changes into the widget, it needs to connect a slot to the setting changed signal. A nice example of this is in `imutils.slideshow` where the slideshow delay is changed according to the setting. All of the logic for setting the metadata text is in the `_update_text` method of the metadata widget where we create and iterate over the `ExifInformation` dictionary updated in #212 . This could probably be re-used / re-factored also for the connected slot.

jeanggi90:
I like you proposition using motions for the shortcut. It make sense to link the *switch set* shortcuts to the *toggle metadata gui* key.

I will have a look what I manage to do and reach out if I need any advice. Thanks a lot for these starter hints!

Issue #214, opened 2020-05-19
Proper exif metadata formatting


karlch:
During #212 we came to the conclusion that proper formatting of the exif metadata is non-trivial. Piexif seems to have no option to give nice correctly formatted information, only the raw data.

We have a few options here:
* Try to get this functionality into piexif. This would be awesome as others could benifit from the implementation, but certainly not guaranteed that upstream wants to support it. Maybe worth trying though.
* Implement a wrapper ourselves and extend this on-demand. We could have a dictionary mapping exif keys to their formatted version, e.g.:
```
formatters = {
    piexif.ExifIFD: {
        piexif.ExifIFD.ExposureTime: lambda value: f"{value:d} s",
    }
}
```
* Switch to another exif tool. This is not my favourite option but could be a last resort. We would probably have to duplicate the current code with both exif dependencies and deprecate piexif for a while. It could be removed for `v1.0.0`. For this we would need to find a much better yet still not too heavy alternative that is really worth it.

Any ideas and suggestions are very welcome!

karlch:
Having some simple proof-of-concept and at least the values of the default keys in the different metadata key-sets properly format before releasing v0.8.0 would make a lot of sense to me. I will probably take a shot at implementing this directly in vimiv and ask you @jcjgraf for some help once the basics are there. Does this sound like a plan to you?

jcjgraf:
I think it definitely makes sense to get the formatting right for the next release.

Doing this formatting ourselves, based on the raw data provided by `piexif` seems a lot of work do me. While it may be reasonable to to it for the most frequently used metadata, it still very tedious. Especially, since the the formatter has to be extended whenever someone needs a new certain new metadata tag.

So my question is, does it really make sense to urgently stick to `piexif` instead of switching to a more elaborated package?
`exiftool` as well as `exiv2` are both amazing tools. While they are not written in python themselves, there exist python wrappers for both.

Obviously both provide formatted metadata output by default (besides many other advantages).

Extracting metadata using `PyExifTool` (wrapper for `exiftool`) is quite simple
```
import exiftool

image = "test.CR2"

with exiftool.ExifTool() as et:
    metadata = et.get_metadata(image)
    
    print(metadata)
```

`metadata` is in this case a dict with all metadata found in the image. The metadata tag is the key and formatted metadata is the value. 

In order to extract only certain tags, one can use `et.get_tags(["firstTag, secondTag"], image)`. which returns a similarly structured dict.

[...]

PR #215, opened 2020-05-21
Mutiple Sets of Metadata

jeanggi90:
As described in #213, this PR implements several customizable sets of metadata keys which are displayed in the `gui.metadatawidget`. The different sets can be accessed and switched by pressing `[count]i`.

- [x] read keysets from the config
- [x] provide a default keyset(s)
- [ ] ~~allow editing via `:set` (???)~~
- [x] react on shurtcut with count and switch set
- [x] better way to define sets in the config. Maye something like `set3 = ...` instead of `3 = ...`
- [ ] ~~*flag* (e.g. `__ALL__`) to display all metadata found, in a certain list~~
- [ ] unit tests (???)
- [x] documentation

The implementation is very similar to how you suggested to implement it in #213. So far, the basic mechanism for reading the keysets from the config, displaying the appropriate metadata in the `gui.metadatawidget` as well as switching the sets using `[count]i` is implemented.  Does it suit your imagination @karlch ?

An open question I have is, how and what default should be provided? Is a single set sufficient (similarly to how it is currently on master)? If yes, I guess we can simply assign that default to `metadata.current_keyset`, since this value should not we changed if `metadata.keysets` is nothing assigned. Further, I am unsure how these sets can be edited during runtime using `:set` and if there should even be this possible. What so you think @karlch ?

P.S. You need to populate you config with some sets in order to test it. Currently it may crash if you do not do so because of the missin default.

### DIFF
--- a/vimiv/api/settings.py
+++ b/vimiv/api/settings.py
@@ -508,4 +508,4 @@ class metadata:  # pylint: disable=invalid-name
         suggestions=defaults,
     )
 
-    keysets: List[str] = defaults
+    keysets: Dict[int, str] = dict(enumerate(defaults, start=1))
--- a/vimiv/api/settings.py
+++ b/vimiv/api/settings.py
@@ -493,12 +493,19 @@ class title:  # pylint: disable=invalid-name
 class metadata:  # pylint: disable=invalid-name
     """Namespace for metadata related settings."""
 
+    # Default sets
+    defaults = [
+        "Make, Model, DateTime, ExposureTime, FNumber, IsoSpeedRatings, FocalLength, LensMake, LensModel, PixelXDimension', 'PixelYDimension, ExposureBiasValue",
+        "ExposureTime, FNumber, IsoSpeedRatings, FocalLength",
+        "Artist, Copyright",
+    ]
+
     # Store the keys as a comma seperated string
     current_keyset = StrSetting(
         "metadata.current_keyset",
-        "",
-        desc="Define the metadata keys to display",
-        hidden=True,
+        defaults[0],
+        desc="Currently displayed metadata keyset",
+        suggestions=defaults,
     )
 
-    keysets: List[str] = None
+    keysets: List[str] = defaults
--- a/vimiv/api/settings.py
+++ b/vimiv/api/settings.py
@@ -495,7 +495,7 @@ class metadata:  # pylint: disable=invalid-name
 
     # Default sets
     defaults = [
-        "Make, Model, DateTime, ExposureTime, FNumber, IsoSpeedRatings, FocalLength, LensMake, LensModel, PixelXDimension', 'PixelYDimension, ExposureBiasValue",  # pylint: disable=line-too-long,useless-suppression
+        "Make, Model, DateTime, ExposureTime, FNumber, IsoSpeedRatings, FocalLength, LensMake, LensModel, ExposureBiasValue",  # pylint: disable=line-too-long,useless-suppression
         "ExposureTime, FNumber, IsoSpeedRatings, FocalLength",
         "Artist, Copyright",
     ]
--- a/vimiv/api/settings.py
+++ b/vimiv/api/settings.py
@@ -499,3 +499,5 @@ class metadata:  # pylint: disable=invalid-name
         "Make, Model, DateTime, ExposureTime, FNumber, isospeedratings, FocalLength",
         desc="Define the metadata keys to display",
     )
+
+    keysets = {}
--- a/vimiv/api/settings.py
+++ b/vimiv/api/settings.py
@@ -495,7 +495,7 @@ class metadata:  # pylint: disable=invalid-name
 
     # Default sets
     defaults = [
-        "Make, Model, DateTime, ExposureTime, FNumber, IsoSpeedRatings, FocalLength, LensMake, LensModel, PixelXDimension', 'PixelYDimension, ExposureBiasValue",
+        "Make, Model, DateTime, ExposureTime, FNumber, IsoSpeedRatings, FocalLength, LensMake, LensModel, PixelXDimension', 'PixelYDimension, ExposureBiasValue",  # pylint: disable=line-too-long,useless-suppression
         "ExposureTime, FNumber, IsoSpeedRatings, FocalLength",
         "Artist, Copyright",
     ]
--- a/vimiv/api/settings.py
+++ b/vimiv/api/settings.py
@@ -494,10 +494,11 @@ class metadata:  # pylint: disable=invalid-name
     """Namespace for metadata related settings."""
 
     # Store the keys as a comma seperated string
-    keys = StrSetting(
-        "metadata.keys",
-        "Make, Model, DateTime, ExposureTime, FNumber, isospeedratings, FocalLength",
+    current_keyset = StrSetting(
+        "metadata.current_keyset",
+        "",
         desc="Define the metadata keys to display",
+        hidden=True,
     )
 
-    keysets = {}
+    keysets: List[str] = None
--- a/vimiv/config/configfile.py
+++ b/vimiv/config/configfile.py
@@ -147,14 +147,14 @@ def _read_plugins(pluginsection):
     plugins.add_plugins(**pluginsection)
 
 
-def _add_metadata(configsetion):
+def _add_metadata(configsection):
     """Set available metadata sets from config file.
 
     Args:
-        configparser: METADATA section in the config file.
+        configsection: METADATA section in the config file.
     """
     temp = {}
-    for name, value in configsetion.items():
+    for name, value in configsection.items():
         match = re.search(r"keys(\d+)", name)
         if match:
             temp[match.group(1)] = value
--- a/vimiv/config/configfile.py
+++ b/vimiv/config/configfile.py
@@ -6,6 +6,8 @@
 
 """Functions to read configurations from config file and update settings."""
 
+import re
+
 import configparser
 
 from vimiv import api, plugins
@@ -153,12 +155,10 @@ def _add_metadata(configsetion):
     """
     temp = {}
     for name, value in configsetion.items():
-        try:
-            # If keyname is int, is a metadata set
-            temp[int(name)] = value
+        match = re.search("keys(\d+)", name)
+        if match:
+            temp[match.group(1)] = value
             _logger.debug("Keyset '%s' found", value)
-        except ValueError:
-            pass
     api.settings.metadata.keysets = [v for _, v in temp.items()]
     if len(api.settings.metadata.keysets) > 0:
         api.settings.metadata.current_keyset.value = api.settings.metadata.keysets[0]
--- a/vimiv/config/configfile.py
+++ b/vimiv/config/configfile.py
@@ -151,10 +151,14 @@ def _add_metadata(configsetion):
     Args:
         configparser: METADATA section in the config file.
     """
+    temp = {}
     for name, value in configsetion.items():
         try:
             # If keyname is int, is a metadata set
-            api.settings.metadata.keysets[int(name)] = value
-            _logger.debug("Updating keyset '%s' to '%s'", name, value)
+            temp[int(name)] = value
+            _logger.debug("Keyset '%s' found", value)
         except ValueError:
             pass
+    api.settings.metadata.keysets = [v for _, v in temp.items()]
+    if len(api.settings.metadata.keysets) > 0:
+        api.settings.metadata.current_keyset.value = api.settings.metadata.keysets[0]
--- a/vimiv/config/configfile.py
+++ b/vimiv/config/configfile.py
@@ -62,6 +62,9 @@ def read(path: str) -> None:
     # Read plugins
     if "PLUGINS" in parser:
         _read_plugins(parser["PLUGINS"])
+    # Read metadata sets
+    if "METADATA" in parser:
+        _add_metadata(parser["METADATA"])
     _logger.debug("Read configuration from '%s'", path)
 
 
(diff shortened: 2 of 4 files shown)


karlch:
Thanks a lot, this looks good!

I would probably provide at least three sets with the default configuration so this feature can be used out of the box. Personally I have absolutely no opinion on the actual values though. I don't think editing the different sets using the `:set` command is needed, one can always set the current `keys` setting to whatever value one likes on the fly with this command anyway. It could be nice to provide completion options for the `keys` setting, e.g. by passing the three sets as suggestions, instead of hiding the setting.

I agree that a nicer name could be useful. `set0` sounds fine, so does `option0` or even just `keys0`. We could then replace the `try ... except` check by a simple regular expression.

I am not sure I understand the `__ALL__` flag, where would the metadata be shown?

It may be difficult to unit test this, but it should certainly be possible to write some basic end2end tests. Let's discuss this after we have agreed on the final implementation. I am also fine with writing these at a later stage, in case you aren't interested in doing these.

Documentation should be as simple as extending the docstring of the metadata command function.

jeanggi90:
>I would probably provide at least three sets with the default configuration[...]

I have added three such sets and linked them the suggestions of `metadata.current_keyset`. These sets are just generalized version of what I find useful and abviously we may change them (esp. since the third list may say empty of the image does not contain any author/copyright information)

>I agree that a nicer name could be useful.

The sets can be defined using the `keys[number]` option. I have used regex, as you had suggested. Since I am not that familiar with Regex in Python, I am open for suggestions if there is a better way to implement this `if` statement.

>I am not sure I understand the `__ALL__` flag, [...]

I thought that it may be useful to provide an easy way to see all metadata saved in the image. By using something like `keys3 = __ALL__`, all metadata of the images would be displayed in list `3` in this case. But I am not sure if this is something which is desired and required.

Concerning the documentations. I think it is really important to document to possible keys to build the metadataset. Unfortunately, piexif to does not provide a complete list in their documentation. Do you think it is sufficient to just say that all possible keys can be found using pythons `dir(piexif....)` command. Or how would you go about that?

Lastly, I am not sure how to fix the remaining linting issue, respectively what causes it to fail. I am glad for any hints :)

karlch:
Thanks again, I am perfectly fine with the defaults you picked. Feel free to change them to anything you feel like could be useful.

I am not sure if displaying all the metadata is very useful as some of them could be random binary blobs, no? What could be nice though is an option to list all of the metadata keys that were found for the current image. This could also serve as an option to find out what keys there are available, instead of relying on piexif documentation. What do you think of this?

Concerning the rest, I will leave a quick code review with in-line comments. This should make things clearer I hope :smile:

vimiv/api/settings.py
```diff
@@ -493,9 +493,19 @@ class title:  # pylint: disable=invalid-name
 class metadata:  # pylint: disable=invalid-name
     """Namespace for metadata related settings."""
 
+    # Default sets
+    defaults = [
+        "Make, Model, DateTime, ExposureTime, FNumber, IsoSpeedRatings, FocalLength, LensMake, LensModel, PixelXDimension', 'PixelYDimension, ExposureBiasValue",
```
karlch, line 498: Feel free to add a `# pylint: disable=line-too-long,useless-suppression` comment here. I suspect that we are hit by https://github.com/PyCQA/pylint/issues/3368 here, thus the additional `useless-suppression`. I don't think splitting such a set into multiple lines makes much sense.

vimiv/api/settings.py
```diff
@@ -493,9 +493,19 @@ class title:  # pylint: disable=invalid-name
...
-        "Make, Model, DateTime, ExposureTime, FNumber, isospeedratings, FocalLength",
-        desc="Define the metadata keys to display",
+    current_keyset = StrSetting(
+        "metadata.current_keyset",
+        defaults[0],
+        desc="Currently displayed metadata keyset",
+        suggestions=defaults,
     )
+
+    keysets: List[str] = defaults
```
karlch, line 511: I would personally prefer a dictionary for this as it:
* Allows the user to keep non-continuous keys, e.g. `1, 2, 3, 8, 9, 0`
* Simplifies keeping defaults and overriding specific options

An option could be:
```
keysets = dict(enumerate(defaults, start=1))
```
where we start with 1 to be consistent with other count-related commands.

vimiv/config/configfile.py
```diff
@@ -6,6 +6,8 @@
 
 """Functions to read configurations from config file and update settings."""
 
+import re
+
```
karlch, line 10: The import order is slightly funny here, I would order them as:
```
import configparser
import re
```

vimiv/config/configfile.py
```diff
@@ -140,3 +145,20 @@ def _read_plugins(pluginsection):
...
     _logger.debug("Plugins in config: %s", quotedjoin(pluginsection))
     plugins.add_plugins(**pluginsection)
+
+
+def _add_metadata(configsection):
+    """Set available metadata sets from config file.
+
+    Args:
+        configsection: METADATA section in the config file.
+    """
```
karlch, line 155: The use of the regular expression is perfect and very clear :+1: 

I see an issue with the current implementation though: This always removes the defaults, no matter if there is anything in the configfile, as the keysets are always overridden. It also doesn't allow keeping some defaults, and overriding others.

With the dictionary as discussed before, this could be simplified somewhat along:
```
if match:
    number = int(match.group(1))
    api.settings.metadata.keysets[number] = value
    _logger.debug("Keyset %d: '%s' found", number, value)
```

vimiv/gui/metadatawidget.py
```diff
@@ -13,6 +13,7 @@
 from vimiv.imutils import exif
 from vimiv.config import styles
 
+from typing import Optional
```
karlch, line 16: This standard library import should go all the way to the top, i.e. here:
```
"""Overlay widget to display image metadata."""

from typing import Optional

from PyQt5.QtCore import Qt
```

vimiv/gui/metadatawidget.py
```diff
@@ -54,20 +55,42 @@ def __init__(self, parent):
             self._mainwindow_bottom = 0
             self._mainwindow_width = 0
             self._path = ""
-            self._loaded = False
+            self._current_set = ""
```
karlch, line 58: I like the logic here a lot! Could you please also update the class docstring `Attributes` section to reflect the change?

vimiv/gui/metadatawidget.py
```diff
@@ -54,20 +55,42 @@ def __init__(self, parent):
...
 
         @api.keybindings.register("i", "metadata", mode=api.modes.IMAGE)
         @api.commands.register(mode=api.modes.IMAGE)
-        def metadata(self):
-            """Toggle display of exif metadata of current image."""
+        def metadata(self, count: Optional[int] = None):
+            """Toggle display of exif metadata of current image and handle switching of metadata keysets.
+
+            **count:** Select the key set to display.
+            """
```
karlch, line 71: Thanks for already working on the docstring! I would remove the "and handle..." from the summary though, and word the count as  "... to display instead.". This keeps it consistent with the slideshow documentation.

In principle this is the docstring I meant before where any user documentation can go. It is accessible via `:help :metadata` and in principle the website. This made me find a bug in the build gh-pages workflow which I have to fix :smile:

vimiv/gui/metadatawidget.py
```diff
@@ -54,20 +55,42 @@ def __init__(self, parent):
...
+
```
karlch, line 72: The implementation here works as I would have imagined, perfect!

I feel like we could simplify this a bit though:
1. Simplify the branching by having an overall `if count is not None: ... elif self.isVisible() ... else: ...`
2. Once again relying on the dictionary, replace the `if self._count_is_valid` with a
```
try:
    # The key set switching logic you implemented
except KeyError:
    # Here I would always raise an error message to the user
    raise api.commands.CommandError(f"Invalid key set option {count}")
```

vimiv/gui/metadatawidget.py
```diff
@@ -54,20 +55,42 @@ def __init__(self, parent):
...
             if self.isVisible():
-                _logger.debug("Hiding widget")
-                self.hide()
+                if count is not None:
+                    if self._count_is_valid(count):
+                        _logger.debug("Switch keyset in open widget")
+                        api.settings.metadata.current_keyset.value = api.settings.metadata.keysets[
+                            count - 1
+                        ]
+                        self._update_text()
```
karlch, line 80: This call to `self._update_text()` should not be needed thanks to the signal connection you did in the constructor.

vimiv/config/configfile.py
jeanggi90: Oh sure, somehow thought `configparser` is a local module :see_no_evil:

vimiv/gui/metadatawidget.py
jeanggi90: Using the `:help` command, I can only access `metadata.current_keyset` but not `:metadata`.

jeanggi90:
I have implemented all your inline-suggestions in commit f67f6f5e611952c13466de65156fc1f4b88fbd0e and 8108b9cfed38a9a349f77ad2840f63cc39a4c007. Concerning the storage of the `keysets` as dict. I have considered using dicts, but thought that lists are easier to use since there are no *gaps* in the count numbering no matter what numbers are chosen for the `keyset` in the config. However, as you have pointed out, this can also be seen as an advantage :)

Anyways, thanks for all you constructive suggestions.

> [...] as some of them could be random binary blobs, no?

I not not think that there is any cryptic binary text, but many values are actually encoded in some numeric way (e.g. ExposureProgram=4 means Shutter Priority mode). This is also something which needs to be addressed in #214.

> What could be nice though is an option to list all of the metadata keys that were found for the current image.

Sounds good. Should this be implemented as a new command, or how did you imagine that?

vimiv/gui/metadatawidget.py
karlch: Is this also the case if you come from image mode?

vimiv/gui/metadatawidget.py
```diff
@@ -54,17 +55,34 @@ def __init__(self, parent):
...
+
+            **count:** Select the key set to display instead.
+            """
+
+            if count is not None:
+                try:
+                    _logger.debug("Switch keyset")
+                    api.settings.metadata.current_keyset.value = api.settings.metadata.keysets[
+                        count
+                    ]
```
karlch, line 78: The linter still complains here, you could split this into two variables:
```
new_keyset = api.settings.metadata.keysets[count]
api.settings.metadata.current_keyset.value = new_keyset
```

vimiv/api/settings.py
```diff
@@ -493,9 +493,19 @@ class title:  # pylint: disable=invalid-name
 class metadata:  # pylint: disable=invalid-name
     """Namespace for metadata related settings."""
 
+    # Default sets
+    defaults = [
+        "Make, Model, DateTime, ExposureTime, FNumber, IsoSpeedRatings, FocalLength, LensMake, LensModel, PixelXDimension', 'PixelYDimension, ExposureBiasValue",  # pylint: disable=line-too-long,useless-suppression
```
karlch, line 498: Are the single quotes around `PixelXDimension` and `PixelYDimension` intentional?

karlch:
Thanks, I find this very clear now. I only left two quick to fix comments I missed in the first iteration, sorry about this! Once these are resolved, I am happy to merge :tada:, unless you want to add end2end tests here. I am fine doing these myself, or if you would like to get into them, which would be awesome of course, having them in an additional PR. Let me know what suits you best.

I deactivated the `if keyname.lower() not in desired_keys` in a quick check which resulted in a crash:
```
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xac in position 22: invalid start byte
```
So I am not sure how to deal with showing everything. Maybe this really does rely on #214 and possibly some skip markers.

I would probably add this as an option to the metadata command again, e.g. `:metadata --list-keys`. Although an additional command is fine as well. But this would certainly go into a new PR, this one has enough complexity on its own :)

vimiv/gui/metadatawidget.py
jeanggi90: Oh I see. Works flawless in image mode

vimiv/api/settings.py
jeanggi90: You are right, they do not make any sense. However, I think I can remove both these key altogether. I expected them to display the resolution of the image, but they do not seem to work.

jeanggi90:
I have applied the two suggested improvements/fixes. Thanks for pointing them out!

Well, I am not that experienced writing unit (/end2end) tests. It is certainly something worth looking into, but currently I do really have the time for that. If you write them I would analyze them, such that I can write them myself for the next PR :innocent:

> I deactivated the `if keyname.lower() not in desired_keys` in a quick check which resulted in a crash:

You are probably right, it is more complex then I expected...

> I would probably add this as an option to the metadata command again

Yeah, it definitely makes more sense to add `list-keys` as an option to the `metadata` command, instead of creating a dedicated command for that. But as you say, lets move this do a seperate PR.

I really like you proposed documentation hint. I have copy/pasted it straight into the docstring and since I do not know the proper formatting of the `hint` I hope I did everything correctly.

If you are happy with all my changes you may merge, if there is something left to do, let me know!

karlch:
Awesome, thanks! You deserve a :beer: or whatever you like to reward yourself :smile:

I like the idea with the end2end tests, I will see when I get to it, probably rather soon given I too have some time to spare.

jeanggi90:
Yeah, preferable something different :relieved:

Thanks to you, I really like contributing to your amazing project!

Concerning the `list-keys` option, shall we open an issue for that so that we do not forget it?

karlch:
Glad to hear all my comments aren't stressing you out :smile:

This certainly makes sense, I opened #220 for this. Feel free to comment on it and let me know if you would also like to tackle this.

PR #217, opened 2020-05-24
Zoom Reset Option

jeanggi90:
I think it is crucial to be able to reset to zoom level back to `fit`, after having to zoom an image e.g. to check the sharpness. Therefore I have added this option, with the default keybinding `=`, for the image and for completeness also the thumbnail view.

### DIFF
--- a/vimiv/commands/argtypes.py
+++ b/vimiv/commands/argtypes.py
@@ -29,6 +29,7 @@ class Zoom(enum.Enum):
 
     In = "in"
     Out = "out"
+    Reset = "reset"
 
 
 class ImageScale(str, enum.Enum):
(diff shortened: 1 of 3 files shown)


karlch:
Thanks a lot for your efforts!

Am I missing something or is this fully equivalent to `:scale --level=fit` which is bound to the `w` key by default? If so, I would be fine with just adding the `=` key as additional default binding for this. What is interesting to see though is that the CI fails as `=` is a valid separator for configparser. We would thus need to treat it specially, much like `<colon>`.

jeanggi90:
You are right, `:scale --level=fit` is the command I was looking for. However, in my view `=` is a more senseful binding. It is actually the default for resetting the zoon for quite a few *vim-like* applications.
I guess in this case you may close this PR, since I can easily change the keywinding in the `keys.conf`.

karlch:
Due to the issue described before, configparser treating `=` as separator, this was not as simple as extending the `keys.conf`. But I agree with you that this is a nice binding to have, so I fixed the configparser issue and added the binding in 641deb2e896abe5b63de3c8d68c72438385ad155 and 876862102e7bc909fbdf53288dce93d0ded9e72f respectively. I hope this is fine with you. Thanks again for pointing this out and working on it!

PR #222, opened 2020-05-29
Extend end2end tests for the :metadata command

karlch:
This PR adds a few more end2end tests for the `:metadata` command.

In addition to the simple visibility checks we now ensure:
* the correct exif information is 
* we can switch exif information using the different key sets and `:[count]metadata`
* invalid `[count]`s trigger the appropriate error message
* the metadata widget updates its content when the image is changed

For this we did a few things:
1. Do some pytest fixture magic to make the `I add exif information` `when` step available for all tests in the image directory. We now use it in both `metadata.feature` and `write.feature`.
2. Extend the exif content in this fixture so we have a bit of information for each metadata key set.
3. Add two more `then` steps to check the text of the metadata widget.
4. Profit by extending the `metadata.feature` with all sorts of tests to ensure the above.

@jeanggi90 feel free to take a look and ask any questions here, happy to answer them if I can! The bdd-like tests using [pytest](https://docs.pytest.org/en/latest/) and [pytest-bdd](https://pytest-bdd.readthedocs.io/en/latest/) certainly take some getting used to but very often we can now write tests with little or no additional code. These end2end tests always run a full blown vimiv process and interact with it, useful for things like testing commands and their effect. For simpler function or class tests we use plain pytest in the unit test folder.

### DIFF
--- a/tests/end2end/features/image/conftest.py
+++ b/tests/end2end/features/image/conftest.py
@@ -1,0 +1,39 @@
+# vim: ft=python fileencoding=utf-8 sw=4 et sts=4
+
+# This file is part of vimiv.
+# Copyright 2017-2020 Christian Karl (karlch) <karlch at protonmail dot com>
+# License: GNU GPL v3, see the "LICENSE" and "AUTHORS" files for details.
+
+import pytest
+import pytest_bdd as bdd
+
+from vimiv import imutils
+
+try:
+    import piexif
+except ImportError:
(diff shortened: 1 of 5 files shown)


jeanggi90:
Thanks for adding these tests! I have never really heard of BDD testing. It's an interesting concept I have to lookup and research myself. I guess the questions will arise when I try to implement them myself for the next PR :blush:

Issue #221, opened 2020-05-29
Display info string if no matching exif keys are found


karlch:
Currently if there are no exif keys in the image matching the keys the metadata widget is just an empty label. This makes it easy to miss and has caught me by surprise a few times now. It could make sense to add a check along:
```
if exif_information:
# as usual
else:
    self.setText("No matching metadata found")
```
or so.

jeanggi90:
I have actually thought about this as well. I think such a comprehensive text may distract a little to much from the remaining valid metadata and maybe a simple hyphen would be more appropriate. Or to do it right, we could add a boolean setting option to enable/disable this behaviour and leave your proposed text.

karlch:
Ah, this is another option I haven't thought about, but I think we have a little misunderstanding. This only concerns the case if there is no metdata matching the current set at all in the current image. For example, if we are at set 3 and the image does not have `Author` or `Copyright`. Currently this leads to the widget being a little empty blob at the bottom right. This could be improved I think.

jeanggi90:
Yeah I have missunderstood you. I have opened a PR concerning this issue: #225

karlch:
Thanks again for your contribution!

PR #225, opened 2020-05-30
Handle Missing Metadata in Widget

jeanggi90:
As discussed in #221 

- [x] Display message when there is not metadata for a given keyset
- [ ] ~~Display `-` for a missing metadata value for a given key (???)~~

I have already implemented the first todo. Concerning the second one (the one I have proposed in #211), I am not sure if this option is really required. Personally, I am fine not showing anything for missing value. But if there is demand for such an option, I can easiliy add it.

implements #221

### DIFF
--- a/tests/end2end/features/image/metadata.feature
+++ b/tests/end2end/features/image/metadata.feature
@@ -48,4 +48,4 @@ Feature: Metadata widget displaying image exif information
         And I run next
         Then the metadata widget should be visible
         # The new image no longer has any exif information
-        And the metadata text should not contain 'Make'
+        And the metadata text should contain 'No matching metadata found'
--- a/vimiv/gui/metadatawidget.py
+++ b/vimiv/gui/metadatawidget.py
@@ -122,7 +122,10 @@ if exif.piexif is not None:
                 "%s: reading exif of %s", self.__class__.__qualname__, self._path
             )
             exif_information = exif.ExifInformation(self._path)
-            self.setText(utils.format_html_table(exif_information.items()))
+            if exif_information:
+                self.setText(utils.format_html_table(exif_information.items()))
+            else:
+                self.setText("No matching metadata found")
             self._update_geometry()
             self._current_set = api.settings.metadata.current_keyset.value
 


karlch:
Thanks again! Would you mind implementing this from the current master to remove the conflicts?

I also see two minor improvements that could be done:
1. You create the `ExifInformation` class twice which reads the exif data twice, this should not be needed.
2. You can just compare to the dictionary directly, an empty dictionary is false, the call to len is not needed.

I also think the second one is not really needed. This can quickly become very verbose.

jeanggi90:
I have missed these commits from you yesterday :grimacing:  I have rebased it and it should by fine now. 

Oh, thanks for pointing out the comparison of dictionaries. There are probably a few python  specific features I gotta learn about.

karlch:
Alright, after the speed of light fix I am happy to merge, yes :+1:

You could however also update the "Switch metadata information upon new image" test in `tests/end2end/features/image/metadata.feature` to check for the new text instead of the missing "Make".

Note that I have noticed some flakiness of these tests on CI that I cannot reproduce locally, so any CI failure containing `None type is not callable` is almost certainly not your fault. Will have to look into this.

jeanggi90:
Yeah I will do that, but first I have to figure out how to run these tests locally... :grin:

karlch:
You will need [tox](https://tox.readthedocs.io/en/latest/). Then something along:
```
tox -e pyqt
```
should work.

You can also look at [the website](https://karlch.github.io/vimiv-qt/documentation/hacking#tests-and-checkers) for a few more hints.

jeanggi90:
I am sorry that I have not been active lately. Currently I am busy. I will finish this PR when I have got some spare time.

karlch:
No worries, life is always more important than this :smile: Let me know in case you cannot finish this and would prefer me to take over, but there is absolutely no hurry.

karlch:
Awesome, thanks!

To me this looks like you don't have `xvfb` installed. In case you are using arch, the corresponding package would be `xorg-server-xvfb`. I noticed that this is not described anywhere in the documentation, shame on me :laughing:

jeanggi90:
That was indeed the problem. Now the tests run successfully

PR #241, opened 2020-06-07
Fix compliance with the xdg trash standard

### LINKED ISSUE #240: Vimiv produces non-compliant .trashinfo file
According to the [freedesktop trash spec](https://specifications.freedesktop.org/trash-spec/trashspec-latest.html):
1. `DeletionDate` must use the `%Y-%m-%dT%H:%M:%S` format
2. Value of `Path` must be urlescaped (e.g. using `quote`/`unquote` from `urllib.parse`).

However, vimiv uses `%Y%m%dT%H%M%S` and does not escape the `Path` value.

karlch:
@woefe could you give this a quick check before I merge?

fixes #240

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -34,6 +34,8 @@ Fixed:
 * Various issues when handling backslash and percent characters in paths and
   completions. Thanks
   `@woefe <https://github.com/woefe>`_ for pointing these out!
+* Quoting of paths and the date format of the trashinfo file created by the ``:delete``
+  command. Thanks `@woefe <https://github.com/woefe>`_ for the bug report.
 
 
 v0.7.0 (2020-05-17)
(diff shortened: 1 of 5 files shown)


woefe:
The completions of `:undelete` still parse the old format. https://github.com/karlch/vimiv-qt/blob/41477003f604d56c648eee7b5c207aed79e318c5/vimiv/completion/completionmodels.py#L203

karlch:
> The completions of `:undelete` still parse the old format.
> 
> https://github.com/karlch/vimiv-qt/blob/41477003f604d56c648eee7b5c207aed79e318c5/vimiv/completion/completionmodels.py#L203

Thanks! This should have been fixed by 166da4cc20ea7e11fc33b0449dea31a9b2cfa8d6.

woefe:
LGTM :+1:

Issue #270, opened 2020-08-30
Key conflicts displayed before parsing the entirety of keys.conf


schyzophrene-asynchrone:
Issue #249 part 2!

Unbinding appropriate keys works, however, since warnings are displayed while parsing `keys.conf`, some of them are still displayed. Suppose I bind `t` before unbinding `ts` for example.

To solve this I could move all `unbind` statements up, but I have the case where I want to bind `s` in `[GLOBAL]`, so I need to unbind `sl`, `sh` and `ss` in `[IMAGE]`. I also have to put the `[IMAGE]` section before the `[GLOBAL]` section. But in `[GLOBAL]` I unbind `k` and bind `kl`, `kh` and `ks` in `[IMAGE]`, so no matter what I do I still have warnings.

The culprit is the `__setitem__` method of the `Trie` class in `vimiv/utils/trie.py`. It should not trigger a warning every time a binding is modified during startup, but wait for `keys.conf` to be totally read.

karlch:
Thanks for the detailed report!

This makes perfect sense and will probably require some refactoring. An option could be to add a `Trie.check` function that triggers any remaining warnings instead of warning instantly in the `__setitem__` as you describe. This would then have to be called
- when `keys.conf` is totally read
- and when calling the `bind` command defined in `config.configcommands.py`.

PR #272, opened 2020-08-30
Checks for binding conflicts only after reading keys.conf or using :bind

schyzophrene-asynchrone:
Since you said where and what to change, I thought I could as well try to do it.

So I made the `Trie.check` you spoke about and removed warnings from `__setitem__`. I’m not totally satisfied with what I came with since we have now only one type of warning (the “hides longer key” type) and from a user perspective, they’re the least useful (when you want to keep the short key but don’t know what the longer one is). But since Trie object don’t have a `parent` attribute, I guess I can’t help it.

This is my first pull request on Github, and I’ve discovered what a digital tree was with your code. My changes are small, but I hope they do what they are supposed to and fit to the general code design you adopted for your project. Any comments are welcome.

#### Related content
Issue #270

### DIFF
--- a/tests/unit/utils/test_trie.py
+++ b/tests/unit/utils/test_trie.py
@@ -43,6 +43,7 @@ def test_setitem_warns_for_hidden_key(trie, mocklogger):
     key1 = "key"
     key2 = key1 + "1"
     trie[key1] = trie[key2] = "value"
+    trie.check()
     mocklogger.warning.assert_called_once()
 
 
@@ -51,6 +52,7 @@ def test_setitem_raises_for_hiding_key(trie, mocklogger):
     key1 = "key"
     key2 = key1 + "1"
     trie[key2] = trie[key1] = "value"
+    trie.check()
     mocklogger.warning.assert_called_once()
 
 
--- a/vimiv/api/keybindings.py
+++ b/vimiv/api/keybindings.py
@@ -158,6 +158,6 @@ def items() -> Iterator[Tuple[modes.Mode, Iterable[Tuple[str, str]]]]:
             yield mode, sort(get(mode))
 
 def check() -> None:
-    """Checks every mode for keybinding clashes and logs warnings"""
+    """Checks every mode for keybinding clashes and logs warnings."""
     for mode, bindings_trie in _registry.items():
         bindings_trie.check()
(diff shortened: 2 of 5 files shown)


karlch:
Awesome, thanks for looking into this! I have a few minor comments and questions, but this looks like a great start, especially for your first PR on Github.

- Looks like you call `api.keybindings.check()` but implemented `utils.trie.Trie.check`. You would need to add another wrapper function in the api. The simplest version would be to just check all modes and associated tries at once:
```
def check() -> None:
    for mode, bindings_trie in _registry.items():
        bindings_trie.check()
```
- If you would like the hidden longer keys in the warning, you could just find them out by iterating over the children. Something along:
```
def check(self) -> None:
    """Checks for possible clashes and logs warnings."""
    if self.key and self.children:
        hidden = []
        for child in self.children.values():
            hidden.extend(key for key, _ in child)
        _logger.warning("'%s' hides longer keys: %s", self.key, quotedjoin(hidden))  # Note: quotedjoin is in utils
```
- The `__setitem__` unit tests in `tests/unit/utils/test_trie.py` related to logging a warning (l. 41 ff. and l. 49 ff.) now fail. This could be fixed by adding a call to your newly implemented `check` method before the assertion. Renaming the tests to reflect that we are testing the check method would be a nice bonus.

Thanks again!

schyzophrene-asynchrone:
So I have added an `api.keybindings.check` function, and have just understood that `api.keybindings` was *not* an instanciation of `utils.Trie`… It’s nice contributing to others code, you get to see things you’re not used to (I never thought of using files like they were classes).

On the “I never thought of that” chapter, next section is about your `Trie` class. I’ve updated the check method as you proposed and as it took me several minutes understanding why this was supposed to work, I was amazed at how powerful this is.

I also have updated the tests, they should pass now.

As you also may see from the commits log, I’m trying to satisfy tox and pass all checks of your repository, but it keeps finding new things to say about my code. That’s definitely a tool I don’t like. :smile:

karlch:
Perfect, thanks a lot!

Using modules/files like classes is one of the things I like in python, although I am still not sure if I am following best-practices here. It certainly feels weird when compared to c++ for example.

I guess I am the one to blame here, not tox which is just used to run all sorts of tests and linters in an isolated environment. The combination of `pylint`, `pycodestyle` and `pydocstyle` is certainly very nit-picky. It gets easier to handle when one just runs the black code-formatter over the codebase with the `pre-commit` hook. Personally, I like the consistent look and sanity checks, although they can be annoying and very noisy. But this is certainly a debatable preference...

So, thanks again, also for fighting through the noise and congrats on getting your first Github PR merged :tada:

karlch:
Sorry, I forgot to mention that you could add yourself to the `AUTHORS` file. Feel free to do this in another quick follow-up PR if you like.

Issue #16, opened 2019-01-06
Support raw images


karlch:
There has been the request for supporting raw images, this should be rather simple according to
https://stackoverflow.com/questions/18637721/how-to-display-a-raw-image-file-in-a-qt-gui.

The file format would have to be detected by the load method in `imutils/imfile_handler.py` and a correct pixmap created from the data.

See [this issue of the GTK version](https://github.com/karlch/vimiv/issues/63).

karlch:
This is actually more difficult than I thought, there seem to be a phletora of different formats and reading any of them directly requires knowing width and height of the image in advance. Possibly simply reading the embedded jpeg is a solution as mention in the issue linked before, but I can only find the small thumbnail version of this which is rather small for displaying. @tobiasbrummer or @ninrod, could you comment on what you mean with the embedded jpeg, is it really simply the thumbnail?

tobiasbrummer:
I don’t know about all cameras as I only have access to my Sony A6300 at the moment. But I just tried to extract the embedded JPG image from a RAW file of the Sony A6300 with „dcraw -e“ and I get a JPG file with ~1600x1080px which I think is enough for preview purposes.

karlch:
Thanks, that was the hint I needed :smile:

With a quick and dirty hack I can get this working. For future reference, something along the lines of
```
output = subprocess.run(["dcraw", "-e", "-c", path], capture_output=True, check=True)
pixmap = QPixmap()
pixmap.loadFromData(output.stdout)
...
```

There are more pythonic solutions, e.g. using [wand](https://github.com/emcconville/wand), but these seem significantly slower as they actually process the complete raw file.

As, in any case, this adds an additional dependency I think this could be a great plugin after #13 is in. Writing the actual API is currently my focus, a raw plugin could then be the next step.

karlch:
I have recently learned about [qt raw plugin](https://gitlab.com/mardy/qtraw). In principle this would provide full support for raw images. Issue with this is that the check we use to determine if a file is an image (`utils.files.is_image`) relies on `imghdr` which has no idea about raw images.

In principle, it is not too difficult to extend  `imghdr` with new formats but every additional check impacts performance. As a compromise, there is now the `imageformats` plugin added in a1b8f421f4c4a2e9538fed9495f41bb5393f9e61. It currently only supports "cr2", as that was the easiest proof-of-concept for me to implement.

It would be great if we could extend `plugins.imageformats` as needed, any help is very welcome :)

[...]

PR #282, opened 2020-10-05
Interface for External Filetype Handler

jeanggi90:
Currently, the plugin `imageformats` fails to open cr2 Raw files. Therefore, I have decided to implement another raw filetype plugin ([see here](https://github.com/jeanggi90/RawPrev)). It follows the idea proposed and discussed in #16 and displays the jpge thumbnail extracted from the Raw.

The make such a plugin work, I had to create a suitable interface within vimiv. My idea was to have a list (`api.external_hander`) to which plugins which want to provide support for additional file types, can add a tuple of the following format: The first value is a list of filetypes which this plugin support. The second value is a callable, which returns an instance of `Pixmap`, containing the image to open.

In `imutils/_file_handler` there is a check which determines whether the type of the current image is in handled by the standard `QImageReader` or by an external handler. In case it is handled externally, the appropriate callable is responsible for creating a `Pixmap` instance with the image to display.

Please @karlch let me know that you think about this idea as well as the implementation. There are certainly many things to improve and maybe there is even a better way to implement this functionality.

**Edit**:
I seem to have troubles on getting the typing for `api.external_handler` right :grimacing:

### DIFF
--- a/vimiv/api/__init__.py
+++ b/vimiv/api/__init__.py
@@ -7,7 +7,8 @@
 """`Utilities to interact with the application`."""
 
 import os
-from typing import List, Iterable
+from typing import List, Iterable, Dict, Callable
+from PyQt5.QtGui import QPixmap
 
 from vimiv.utils import files
 
@@ -27,6 +28,7 @@ from vimiv.api import (
 )
 
 mark = _mark.Mark()
+external_handler: Dict[str, Callable[[str], QPixmap]] = {}
 
 
 def current_path(mode: modes.Mode = None) -> str:
--- a/vimiv/imutils/_file_handler.py
+++ b/vimiv/imutils/_file_handler.py
@@ -120,13 +120,22 @@ class ImageFileHandler(QObject):
             api.signals.movie_loaded.emit(movie, reload_only)
             self._edit_handler.clear()
         # Regular image
-        else:
+        elif file_format in reader.supportedImageFormats():
             pixmap = QPixmap.fromImageReader(reader)
             if reader.error():
                 log.error("Error reading image %s: %s", path, reader.errorString())
                 return
             self._edit_handler.pixmap = pixmap
             api.signals.pixmap_loaded.emit(pixmap, reload_only)
+        # Externally handled image
+        elif file_format in api.external_handler:
+            handler = api.external_handler[file_format]
+            pixmap = handler(path)
+            self._edit_handler.pixmap = pixmap
+            api.signals.pixmap_loaded.emit(pixmap, reload_only)
+        else:
+            log.error("Error reading image %s: fileformat not supported", path)
+            return
         self._path = path
 
     @api.commands.register(mode=api.modes.IMAGE)
--- a/vimiv/utils/files.py
+++ b/vimiv/utils/files.py
@@ -13,6 +13,8 @@ from typing import List, Tuple, Optional, BinaryIO, Iterable, Callable
 
 from PyQt5.QtGui import QImageReader
 
+from vimiv import api
+
 
 def listdir(directory: str, show_hidden: bool = False) -> List[str]:
     """Wrapper around os.listdir.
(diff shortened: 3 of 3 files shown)


karlch:
Thanks for the PR!
I am unfortunately rather busy this week but I hope to get to some vimiv development over the week-end.

karlch:
I like the idea a lot. I was actually thinking of something similar some time ago, but this is much better and nicer than anything I came up with, thanks a lot!

The only thing I would change is to simplify the type of `api.external_handler`. How about removing one level of nested types and just store everything in a dict that maps individual file formats to the appropriate handler, like so:
```python
external_handler: Dict[str, Callable[[str], QPixmap]] = {}
```

In case the same handler can be used for different types, it would have to be registered for all of the types in the dictionary. But this seems like a small caveat compared to how much simpler the code becomes in other places. E.g. in `imutils._file_handler` we could do something like
```python
        elif file_format in api.external_handler:
            handler = api.external_handler[file_format]
            pixmap = handler(path)
            self._edit_handler.pixmap = pixmap
            api.signals.pixmap_loaded.emit(pixmap, reload_only)
        else:
            log.error("Error reading image %s: fileformat not supported", path)
```
and the test in `utils.files` just becomes
```python
            if (
                name in QImageReader.supportedImageFormats()
                or name in api.external_handler
            ):
                setattr(test, "checked", True)
                return name
```

What do you think?

karlch:
I noticed that I just pushed a change which makes merging this non-trivial. For me the code looks fine and all tests pass locally without any issues. Unless you have something else you would like to add, I would merge via the command-line and push.

jeanggi90:
I am glad you like the idea. Yeah, I have just pushed you suggestion, which I thinks, makes the code indeed much cleaner!

> I noticed that I just pushed a change which makes merging this non-trivial.
Am not quire sure that you mean. Should I have rather pushed instead of force-pushed?

I am fine, there is nothing I want to add. The only thing I am unsure is, if the `if reader.error()` (line 125) conditional in `vimiv/imutils/_file_handler.py` is still required (see my code comment in GH).

karlch:
Awesome, thanks!

No, your force-pushing is perfectly fine, I am the one to blame here. I had refactored the `_file_handler` a while ago as I noticed an inconsistency between how it is done for images and thumbnails. I did not push this change until a few minutes ago though. This makes your perfectly fine changes incompatible with my havoc. I will sort out my mess :laughing:

jeanggi90:
Great! 

Wait a second! In the thumbnail view no images are displayed. I noticed that a few days ago and have not looked further into it. It just came to my minde when you were talking about this inconsistancy between images and thumbnails.

karlch:
Yes, this makes perfect sense, in an ideal world this would be auto-solved by me sorting out the chaos, let's see :laughing:

jeanggi90:
Haha, I see. Good luck!

I noticed that thumbnail problem when I was working on this interface, but since I was dealing with other problems at that time I forgot to have a look at it or take a note :see_no_evil: 

Anyways, let me know if there is anything I could help you out with.

karlch:
So I came up with something, but I am not 100 % happy with the solution. Another option I had thought of is to:
- expose `utils.imagereader.BaseReader` via the api
- add a utility function, like `api.add_imageformat(file_format: str, reader: BaseReader, check)` which would take care of everything.

Plugins, like your nice raw handler, would then have to:
- inherit the reader from `BaseReader` and implement `supports` as well as `get_pixmap`
- define the check for `imghdr`
- pass everything to the `api`

This would then remove the need of the `ExternalReader` in `utils.imagereader`. Instead, the `READERS` would be extended by user-defined reader classes that define which formats they support.

This is just random brainstorming, what do you think about this?

Note: To get thumbnail support for raw images, you will probably have to delete your thumbnail cache.

jeanggi90:
Ah, now I understand the problem you were facing :blush: 

I think you implementation is very slick and clean. `BaseReader` builds a nice interface which wraps all image related functionality required to open and display images.

To be honest, I am not really satisfied with the current API for external readers. You have build such a handy `BaseReader` class, and now you hide it from the API by using my inflexible, error-prone `external_handler` dictionary. :smile: 
As you have said, we should really expose the `BaseReader` to the API and ad functionality to push custom `Readers`. It is a much cleaner interface for image format plugins. Additionally, it is also much easier to extend it, in case we need to expose other functionality, which may be required for certain image types, in the future.

I see, I have created my PR at the wrong time :upside_down_face:

PR #288, opened 2020-10-30
API convenience function for adding filetype support

jeanggi90:
To finish up #282 and a655a71cd61ffce6d9951640b474833b95562131 I have added a convenience function to the API which allows to add support for new file types using:
```
api.add_external_format(fmt="cr2", test_func=test_cr2, load_func=load_cr2)
```
instead of:
```
files.add_image_format("cr2", test_cr2)
api.external_handler["cr2"] = load_cr2
api.external_handler["CR2"] = load_cr2
```

Further, I have moved the `external_handler` dictionary from `vimiv/api/__init__` to `vimiv/utils/imagereader` since there is no need for it to be exposed via the api. @karlch if you think there is a better place for the `external_handler` dict let me know.

### DIFF
--- a/vimiv/api/__init__.py
+++ b/vimiv/api/__init__.py
@@ -7,10 +7,10 @@
 """`Utilities to interact with the application`."""
 
 import os
-from typing import List, Iterable, Dict, Callable
+from typing import List, Iterable, Callable, BinaryIO
 from PyQt5.QtGui import QPixmap
 
-from vimiv.utils import files
+from vimiv.utils import files, imagereader
 
 from vimiv.api import (
     commands,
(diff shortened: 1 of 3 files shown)


karlch:
Awesome as always, thanks! I will fix the `lint` after merging.

jeanggi90:
Sorry for that! Locally, `tox` reported that everything is fine and GitHub did not let me display the error log:
```
Error: We are currently unable to download the log. Please try again later. 
```

karlch:
No worries, there also seems to be some flakiness in the test CI currently, I will try to investigate but cannot reproduce locally.

PR #298, opened 2020-11-08
Add --keep-zoom flag for :next and :prev

karlch:
Adds a `--keep-zoom` flag to the `:next` and `:prev` command which preserves the zoom level and scroll position of the current image.

This was very easy to implement as all of the logic was already there, just hidden in form of a `reload_only` argument.
@jcjgraf would you like to give this a shot before I merge? Hope you don't mind me dashing ahead here, as you can see the code for this was more or less already in-place.


jcjgraf:
This works like a charm, thanks a lot!

PR #301, opened 2020-11-17
Read Metadata Using py3exiv2

jcjgraf:
Depreciating `piexif` in favour of  `py3exiv2` to allow, besided others, out-of-the-box value formatting. Discussed in #214.

You @karlch have suggested to create an additional `imutils.exif_wrapper` wrapper module. You have to elaborate that a bit more. Currently, the new `py3exiv2` as well als the old `piexif` code is written directly in `imutils.exif`. My idea of putting all new code directly into `imutils.exif` was that once `piexif` is depreciated, we can simply remove all `piexif` related code from `imutils.exif`. This way we end up having all exif related code in `imutils.exif`, at is it currently.

Some things to consider:
- Is there a specific reason why the functions `copy_exif` and `exif_do_time` are not class methods of `ExifInformation`? This way there are two independent dependency checks, once in the `@check_pyexiv2` decorator and once inside the initialiser of `ExifInformation`. Would it not make more sense to add these two functions as getters to the `ExifInformation` class?
- `Exiv2`has the convention to refer metadata fields using a key of the format `familyName.groupName.tagName`, while `piexif` simply uses `tagName`. How to maintain backward compatibility?
- Change unit tests to only test `py3exiv2` or still test `piexif` too?

This PR is still very WIP.


karlch:
Awesome, thanks a lot! I will take a closer look at the code tomorrow, but can give some quick answers to the open topics:

- I wanted to avoid the code in `exif` should not become too crowded with the wrapper functions around the two exif libraries used. But looking at the code, this does not seem to bad and I agree we can stick to having everything in there.
- Both functions don't really need the full functionality of the `ExifInformation` class, so I wanted to avoid having to create one and parse/format the exif information in them. Especially copying the exif metadata to a file seems like a very different operation than creating a nicely formatted metadata dictionary, so I don't feel like they should be grouped together.
- Could we also just use `tagName` now? The full key seems wildly verbose, especially when used in the config, no?
- Ideally we would have the tests for both libraries currently, although this will probably require some funky patching. But we can probably add a pytest fixture that sets the framework and possibly parametrize the tests in the framework.

jcjgraf:
- While both functions do certainly not requires the extraction of formatted metadata, they still have in common the dependency check as well as the loading of the metadata using `pyexiv2/piexif`. In 3ff7a7132bceec5e9ba5dbe2065d64eaa25cc1f7 I have made a proposition to make them class methods. To prevents the loading of the formatted metadata (call to `_load_exif`) this methods would have to be "lazily loaded" (I would have to research how this is best done in Python). Ignoring the burden of supporting the depreciated `piexif`, I think this would be a quite clean solution. But we may also revert this commit if you prefers to have these two functions separate.
- In `piexif` tags were actually also ordered hierarchically in some way. But what we did was we iterate over all tags and put them into a single dict. This could theoretically lead to `tagName` clashes I guess. Give, this nicely ordered structure, I opt for not getting rid of it but use it for our advantage. But I agree that listing these tags may clutter up the config. Here some ideas:
    - Allow something like `familyName.groupName{tagName1, tagName2, tagName3}`
    - Given simply `tagName`, assume it to belong to `Exif.Image` or `Exif.Photo` (which are the most "commonly" used exif groups).
- I also think it makes sense to test both both libraries as I cannot guarantee that I have not broken anything `piexif` related :upside_down_face: 

Just as a side note. Even though everyone talks about solely *Exif* (e.g. tools are named `exiv2`, `piexif`, `exiftools` etc.), this is actually only one of many image metadata standards. It is the one containing information about camera settings and about the image (resolution, format etc.). But there are also IPTC (important for copyright, keywords etc.), XMP as well as some lesser known standards. However, libraries like `pyexiv2` can deal with all common standards (though, our current implementation does not support XMP).

karlch:
I had a first look at the code but suspect that you are still probably want to play around some more before an actual code review. Let me know in case this is wrong or whenever you think you are ready :relaxed:

- In case we go with this, I would recommend not inheriting `ExifInformation` from `dict` and renaming it into `ExifHandler` or so as the class is now more than just an info dict. We could then replace the `_load_exif` with a method like `get_formatted_exif` or so which returns a dictionary instead of loading it into the class. Then we might have the best of both worlds, handle the initial loading and dependency check in the constructor, do everything else in dedicated methods. What do you think?
- I have no strong opinion on this, although it might be more difficult to build together the name for `piexif` compared to dropping the first bits from the `pyexiv2` key name. In case this is not an issue, we can go for the long name with the "belongs to" assumption you describe. I am not sure if the fancy grouping syntax is worth the extra effort when parsing the config though, as always this can wait for another issue/PR if you think otherwise.
- Sounds good, seems like quite a few of the exif related tests are currently failing anyway :blush: Let me know if I can be of any help with the test havoc :laughing:

jcjgraf:
You were right, the code was not ready for detailed review. But sometimes early feedback useful to make sure I do not head in the completely wrong direction :blush: 
But now, the implementation is pretty much complete (except for the tests). So, if you have some spare time it would be great if you could have a look at the code (the commits are quite a mess, I know. I will clean them up before the final merging)

- I have learned that decorators can be used to classes to. So what I have implemented is:
    - `ExifHandler` (the renamed `ExifInformation` as you suggested) contains all functionality to work with the metadata under the assumption that `py3exiv2` is available. If it was not, the plain class would crash.
    - The `ExifHandler` is decorated by `check_exif_dependancy`. This function returns the original `ExifHandler` in case `py3exiv2` is available. If it is not, it is checked if `piexif` is available. If so, a subclass of `ExifHandler` is returned which does all the work based on `piexif`. If none of the two modules is available, a fallback implementation `NoExifHandler` is returned which logs on all methods accesses that no exif module is available. But beside debug logs, the method calls should also report that there is no exif support by writing a warning message to the status line (not implemented yet).
- Something I have noticed is that when there is no exif support, the code following line 21 in `vimiv/gui/metadatawidget` is not executed. Therefore, the `metadata` command is not registered and hence, nothing is logged when the user tries to open the metadata. In my opinion, the user should get a feedback that this functionality is only available with a exif library installed (e.g. by forwarding the request to the fallback class `NoExifHandler` as described above).

As the ExifHandler implementation is pretty much complete, we gotta discuss how I should implement the tests. I am not really experienced with `pytest` so I guess you gotta give me some hints :upside_down_face: 

The testing environment is (obviously) designed to be used with `piexif`. But since `py3exiv2` seems to be available in the environment too (correct me if I am wrong), the test are executed using `py3exiv2`. Even though I am not sure why, this seems to be the cause why some tests are currently failing.

karlch:
Awesome work, thanks a lot! I will take a detailed look into this over the week-end :blush:

vimiv/gui/metadatawidget.py
```diff
@@ -18,7 +18,7 @@
 _logger = utils.log.module_logger(__name__)
 
 
-if exif.piexif is not None:
+if exif.piexif is not None or exif.pyexiv2 is not None:
```
karlch, line 21: Maybe we could add something like `exif.has_exif_support` which is currently simply `pyexiv2 is not None or piexif is not None` so we don't have to know about the library details elsewhere.

vimiv/imutils/exif.py
```diff
@@ -11,138 +11,267 @@
...
-    logged. It it is available, the function is called as usual.
+    If py3exiv2 is available the class is left as it is. If py3exiv2 is not available
+    but the depreciated piexif module is, a depreciation warning is given to the user
+    and a ExifHandler class supporting piexif is returned. If none of the two modules
+    is available, a dummy class is returned and a debug log is logged.
 
     Args:
-        return_value: Value to return if piexif is not available.
+        return_value: Value to return if neither py3exiv2 nor piexif is available.
+        check_piexif: Check piexif dependency too in case py3exiv2 is not available.
```
karlch, line 34: Is there any reason to have `check_piexif` as a decorator argument? For testing?
Otherwise we can just remove the piexif part as soon as piexif support is removed, i.e. before `v.1.0.0`.

vimiv/imutils/exif.py
```diff
@@ -11,138 +11,267 @@
...
-            _logger.debug(
-                "Cannot call '%s', piexif is required for exif support", func.__name__
-            )
+    def decorator(handler):
+        if pyexiv2:
+            return handler
+
+        if piexif and check_piexif:
+
+            class ExifHandlerPiexif(handler):
```
karlch, line 43: I have a few questions here:
- Why does this inherit from the other handler? As far as I can tell, there is no shared code.
- Could we just define this class (as well as `NoExifHandler`) outside of the decorator, i.e. below the regular handler,  then to avoid this huge nesting?

vimiv/imutils/exif.py
```diff
@@ -11,138 +11,267 @@
...
+                    get_formatted_exif: Get dict containing formatted exif values.
+                    copy_exif: Copies the metadata to the src image.
+                    exif_date_time: Get the datetime.
+
+                Attributes:
+                    _metadata: Instance of the pyexiv2 metadata handler
+                """
+
+                def __init__(self, filename=""):
+                    super()
```
karlch, line 59: This call seems a bit superfluous :smile:

vimiv/imutils/exif.py
```diff
@@ -11,138 +11,267 @@
...
+
+            return ExifHandlerPiexif
+
+        if return_value is not None:
             return return_value
 
-        if piexif is None:
-            return stub
-        return func
+        class NoExifHandler:
```
karlch, line 168: As mentioned above, I think it would be nice to define the class outside of the decorator to avoid the nesting.

vimiv/imutils/exif.py
```diff
@@ -11,138 +11,267 @@
...
+            """Handler for exif access without hander library."""
 
-    return decorator
+            def __init__(self, *args):
+                _logger.debug("No exif support.")
 
+            def __getattribute__(self, name):
```
karlch, line 174: As much as I like avoiding code duplication, currently this does not work properly as the `lambda` always returns `None`. This crashes on `exif_date_time` with a `TypeError` as we expect a string, here the empty string, as return value.

I guess `get_formatted_exif` would die in a similar fashion as we always expect a `dict`, but this is currently not called anywhere without exif support.

I see two options here:
- Implement the functions explicitly and return the proper type after logging / messaging.
- Make the lambda raise an exception, ideally a new custom exception like `NoExifSupport`, and enforce the caller to deal with it explicitly instead of checking for `... is None ...`. Like this the caller can think about messaging or the return type. The code in `filelist.exif_date_time` would then read:
```
@api.status.module("{exif-date-time}")
def exif_date_time() -> str:
    """Exif creation date and time of the current image.

    This is meant as an example api.status.module to show how to display exif
    data in the statusbar. If there are any requests/ideas for more, this can
    be used as basis to work with.
    """
    try:
        return imutils.exif.ExifHandler(current()).exif_date_time()
    except imutils.exif.NoExifSupport:
        return ""
```
and this `__getattribute__` could be
```
            def __getattribute__(self, name):
                if name in ["copy_exif", "exif_date_time", "get_formatted_exif"]:
                    raise NoExifSupport()
                return super().__getattribute__(name)
```

I somehow feel like the exception version could be the nicer, and more extendible, version. What do you think?

vimiv/imutils/filelist.py
```diff
@@ -122,7 +122,7 @@ def exif_date_time() -> str:
     data in the statusbar. If there are any requests/ideas for more, this can
     be used as basis to work with.
     """
-    return imutils.exif.exif_date_time(current())
+    return imutils.exif.ExifHandler(current()).exif_date_time()
```
karlch, line 125: This could possibly be changed according to the comment above if we opt for the exception version.

tests/unit/imutils/test_exif.py
```diff
@@ -11,17 +11,36 @@
 from vimiv.imutils import exif
```
karlch, line 11: I need some more time to think about the testing in general, so this review will not cover the tests yet :smile:

karlch:
So besides the inline comments, some more thoughts for general discussion :blush:

Personally I don't mind having lot's of commits, this shows the thought history, but of course as this is your PR it is up to you how you would like to keep them organized.

- I really like the idea of using a class decorator for this :+1:
- Personally I like the behaviour that the `:metadata` command is not registered when there is no exif support as e.g. commands also show up in the completion, and having completion for a command that just displays a warning seems rather unintuitive to me. The behaviour of the code in case there is no exif support is still up for discussion, as shown by the question concerning raising `NoExifSupport` instead of the current behaviour. I am also not completely sure what to log when yet, or which messages to show the user when. This probably also depends on the occasion, some random thoughts in the next bullet points.

Logging with piexif:
- Warn once upon startup that piexif is deprecated, this could happen right in your fancy class decorator.

Logging when there is no metadata support:
1. `:metadata`: Personally I am happy with not defining it at all, I would certainly want to keep it out of the completion. If you feel like there should be a warning when running `:metadata` explicitly, although I don't know why this would happen except for maybe user-defined bindings that come from a system with metadata support, we could define a stub command that is hidden from completion in the `... else` part of `gui.metadatawidget`:
```
else:  # No exif support  # pragma: no cover  # Covered in another CI
    MetadataWidget = None  # type: ignore

    @api.commands.register(mode=api.modes.IMAGE, hide=True)
    def metadata():  # We may want to mirror the arguments
        """Dummy metadata command to display a warning in case explicitly called."""
        raise api.commands.CommandWarning("Please install py3exiv2 for exif support")
```
2. `exif-date-time` status module: log a warning when first called, not every time
3. `copy_exif`: this is the most difficult and I don't really have an opinion here yet :laughing: I guess we could warn on every write, maybe we could even ask the user for explicit confirmation.

vimiv/gui/metadatawidget.py
jcjgraf: I have actually considered replacing this check with something like `if exif.ExifHandler() != "":` That was actually the main reason why I have added the `filename=""` default. But in the end I did not add this because this check required deeper knowledge of the `ExifHandler` implementation.

Probably something as you suggest is most senseful.

vimiv/imutils/exif.py
jcjgraf: I have added this only for the tests in `tests/unit/imutils/test_exif`. We can also get rid of this option and design the tests differently, no problem.

vimiv/imutils/exif.py
jcjgraf: - There is theoretically not need for inheritance, since everything gets overwritten. The idea of this implementation was to get some kind of a *cheap* interface/abstract class (indicate that the both classes have the same functionality). But it is probably better to remove the inheritance completely or do it properly using abstract classes.

vimiv/imutils/exif.py
jcjgraf: I see, my idea was probably not yet well-conceived :D I think both of your ideas are reasonable. If we implement the functions explicitly, we could use an abstract class the to make it more obvious that all three classes have the same functionality, as mentioned above. I also quite like the idea of of throwing a custom exception. Probably, this would be the more pyhtonic solutions.

jcjgraf:
Thanks for your thoughts and ideas! I do agree with your points concerning logging.

- It certainly does not make sense and is probably rather confusing for users to registering commands which are not supported. But for new users who have not realized that an optional package is required for exif support, it may be equally confusing if nothing happens even though pressing `i` should show the metadata (aka me a few months ago :upside_down_face: ). But we may also leave it as it is or if desired, add this functionality in a different PR.
- In my view it is sufficient to warn the user that without exif support, images are saved without additional exif information. We could certainly try to extract the exif from the binary image and try to dump it into the new image or something like that. But isn't the purpose of packages exactly to get this kind of modularity?

A question concerning logging. Is there some built-in functionality to make sure that a certain log is only logged once or how is this done best (e.g. for `exif-date-time`)?

vimiv/imutils/exif.py
jcjgraf: Improved in 4fd97f9f4bdb73e583092ab49486144d4a08fb36

vimiv/imutils/exif.py
jcjgraf: I have decided to use a abstract class `_ExifHandler` and define three implementations of this class. This leas to more code compared to the  `__getattribute__` based implementations but it seems to be cleaner and less *hacked*. But to get the best of both sides, I have also added the `NoExifSupport` custom exception.

Implemented in 4fd97f9f4bdb73e583092ab49486144d4a08fb36 and 4fd97f9f4bdb73e583092ab49486144d4a08fb36.

vimiv/imutils/exif.py
karlch: Much nicer, thanks! If defining another class ever becomes a relevant performance issue, we have done a lot of other things right :laughing:

vimiv/imutils/exif.py
karlch: Although I have no issue with a funky `__getattribute__` implementation this is certainly cleaner, nice!

vimiv/gui/metadatawidget.py
karlch: Looking good!

vimiv/imutils/exif.py
karlch: I think we can get around both of these arguments by making smart use of pytest fixtures. I would assume that we currently:
- always want to check for piexif
- always want to return one of the three implementations of the `ExifHandler` class, never `None` or another funky return value

vimiv/imutils/exif.py
```diff
@@ -10,139 +10,269 @@
...
+            e.strip() for e in api.settings.metadata.current_keyset.value.split(",")
+        ]
+        _logger.debug(f"Read metadata.current_keys {desired_keys}")
+
+        exif = dict()
+
+        for key in desired_keys:
+            # For backwards compability, ssume it has one of the following prefixes
+            for prefix in ["", "Exif.Image.", "Exif.Photo."]:
+                key = f"{prefix}{key}"
```
karlch, line 236: We need something like
```
        for base_key in desired_keys:
            # For backwards compability, ssume it has one of the following prefixes
            for prefix in ["", "Exif.Image.", "Exif.Photo."]:
                key = f"{prefix}{base_key}"
```
otherwise we get `Exif.Image.Exif.Photo.Whatever`. That is a nasty little bug that causes the one test to fail as the `ExposureTime` key is not recognized.

vimiv/imutils/exif.py
```diff
@@ -10,139 +10,269 @@
...
+
+        return exif
+
+    def copy_exif(self, dest: str, reset_orientation: bool = True) -> None:
+
+        if reset_orientation:
+            with contextlib.suppress(KeyError):
+                self._metadata["Exif.Image.Orientation"] = ExifOrientation.Normal
+        try:
+            self._metadata.copy(dest)
```
karlch, line 262: This crashes when writing. I assume that the py3exiv2 method requires `dest` to be an instance of `pyexiv2.metadata.ImageMetadata` and not simply `str` like we had for piexif.

karlch:
Concerning the logging thoughts:
- How about we just warn once upon startup, using the module logger of the exif module within the fancy class decorator, that there is no exif support and therefore:
    - Exif data is lost when writing images to disk
    - The `:metadata` command and associated `i` keybinding is not available
    - The `{exif-date-time}` statusbar module is not available
- Keep everything else to a later PR. Implementing something that only logs messages once is not very difficult, but I feel like it shouldn't be mixed with this already complex PR :blush:

vimiv/imutils/exif.py
jcjgraf: Oh silly me, good catch!

jcjgraf: Fixed in 124fb46fbd3343d8209225d4c20c36771a87a764

vimiv/imutils/exif.py
jcjgraf: That is true. I should have read the docs more closely. Thanks a lot.

Fixed in 1ea3f98df288fc23ec1507933c96bef65f36830b

vimiv/imutils/exif.py
```diff
@@ -10,139 +10,271 @@
...
+            piexif.InvalidImageDataError, FileNotFoundError, KeyError
+        ):
+            return self._metadata["0th"][piexif.ImageIFD.DateTime].decode()
+        return ""
+
+
+class _ExifHandlerNoExif(_ExifHandler):
+    """ExifHandler implementation for no exif support."""
+
+    def __init__(self, filename=""):
```
jcjgraf, line 160: I get a linting error here because `filename` is unused. Can I simple add a `# pylint: disable=unused-argument` to ignore it it is there a better solution?

vimiv/imutils/exif.py
jcjgraf: Improved in bc5fe863a3641910b8bc3793b2a60e5fe2c8289c

vimiv/imutils/exif.py
karlch: You could also just prepend it with an underscore, i.e. `_filename`, to indicate that it is unused.

jcjgraf: Thanks, fixed in c3feef5f119e2ebd3190ce87d7e0d567745429c5

jcjgraf:
I am sorry for the rather slow progress I am making. In 14f113bc3cd38167f4810ce4d3b6cc478dd96e6b I have added the logging as you have suggested it in the previous comment.
Currently, two of the testcases are still failing and I am not quite sure why. Do you have an idea what it going wrong?

jcjgraf:
Just to not forget about it I am posting this here. While randomly browsing some raw images (only looking at some images, not write or anything) I have encountered a crash. Here the traceback:

```
[17:58:11] ERROR    <vimiv>              Uncaught exception! Exiting gracefully and printing stack...
Traceback (most recent call last):
  File "/home/jeanclaude/Documents/Programming/vimiv-qt/vimiv/utils/__init__.py", line 240, in run
    self.function(*self.args, **self.kwargs)
  File "/home/jeanclaude/Documents/Programming/vimiv-qt/vimiv/imutils/_file_handler.py", line 179, in write_pixmap
    _write(pixmap, path, original_path)
  File "/home/jeanclaude/Documents/Programming/vimiv-qt/vimiv/imutils/_file_handler.py", line 214, in _write
    imutils.exif.ExifHandler(original_path).copy_exif(filename)
  File "/home/jeanclaude/Documents/Programming/vimiv-qt/vimiv/imutils/exif.py", line 263, in copy_exif
    dest_image.read()
  File "/home/jeanclaude/Documents/Programming/vimiv-qt/.venv/lib/python3.9/site-packages/pyexiv2/metadata.py", line 115, in read
    self.__image = self._instantiate_image(self.filename)
  File "/home/jeanclaude/Documents/Programming/vimiv-qt/.venv/lib/python3.9/site-packages/pyexiv2/metadata.py", line 87, in _instantiate_image
    return libexiv2python._Image(filename)
OSError: Failed to read input data
```
Also something just for the record, the previous image comment problem with TIFF images actually carries over in an extended way to different formats. So does for example CR2 support everything but image comments while CRW supports image comments but not XMP and ICC...

jcjgraf:
In b4724c8b3a8ff8143df8128a581ff0e48df9aaba I have fixed the second error I had reported in the previous comment. Unfortunately, the check for legal arguments of `pyeciv2.ImageMetadata.copy`is not done by the library but has to be done manually. @karlch if you have an idea for a nicer implementation of this let me know.

karlch:
Thanks for all your comments and hard work on this rather tedious matter. Absolutely no worries concerning the slow progress, my comments are also rather late. I am starting to take a closer look and hope to comment with more details during this week, depending on how intense it is. At the very latest I will have quite some time and probably also motivation, as there is no work-related programming, over the Christmas holidays.

vimiv/imutils/exif.py
```diff
@@ -10,139 +10,361 @@
...
+
+        if reset_orientation:
+            with contextlib.suppress(KeyError):
+                self._metadata["Exif.Image.Orientation"] = ExifOrientation.Normal
+        try:
+            dest_image = pyexiv2.ImageMetadata(dest)
+            dest_image.read()
+
+            self._metadata.copy(dest_image, *get_copy_args())
+
```
karlch, line 354: Adding
```
dest_image.write()
```
after the copying of the metadata fixes the test suite for me. Obviously the manual check for the file-types is rather broken, I will take some more time to look into this :smile:

jcjgraf: How could I miss something so obvious? I have expected that the pyexif based test is somehow broken. :laughing:  Thanks for the hint!

vimiv/imutils/exif.py
```diff
@@ -10,139 +10,361 @@
...
+            return tuple(args)
+
+        if reset_orientation:
+            with contextlib.suppress(KeyError):
+                self._metadata["Exif.Image.Orientation"] = ExifOrientation.Normal
+        try:
+            dest_image = pyexiv2.ImageMetadata(dest)
+            dest_image.read()
+
+            self._metadata.copy(dest_image, *get_copy_args())
```
karlch, line 353: I am a bit afraid of the behemoth I just came up with, but how about not checking the file type but just trying the different parts one-by-one and ignoring any errors (we could add a debug log instead of just suppressing if we opt for this):
```
    dest_image = pyexiv2.ImageMetadata(dest)
    dest_image.read()
    for copy_args in set(itertools.permutations((True, False, False, False))):
        with contextlib.suppress(ValueError):
            src_image.copy(dest_image, *copy_args)
    with contextlib.suppress(OSError):
        dest_image.write()
```

This seems rather pythonic as we ask for forgiveness instead of permission, but I am not happy with the fact that the pyexiv2 does not seem to provide a simple method to get the copy args we are manually creating/iterating.

vimiv/imutils/exif.py
jcjgraf: Fixed in 9f6acd8917ae9e20932098cdac2ea3402b4fa673

vimiv/imutils/exif.py
jcjgraf: Oh wow, your solution is actually really smart. At first I did not get it and I already wanted to comment that this is not gonna work :blush:  But it is probably the clearest solution we can get.

It is indeed a bit unsatisfying that `py3exiv2` does not provide a more handy interface for copying metadata. But since this package is maintained, we could try to get this functionality into `py3exiv2` maybe.

karlch: First of all, merry Christmas Jean-Claude, I hope you enjoy your holidays :christmas_tree:

Would you like to add this in then? Probably with some debug logging like:
```
try:
    ...
except OSError as e:
    _logger.debug("Failed to write exif data for '%s': ''%s'", dest, str(e)")
```

tests/unit/imutils/test_exif.py
karlch: If you don't mind, I feel like it would be easier for both of us to try to get the CI passing here and merge this PR afterwards. I will likely have to play around a lot to see how to test all three scenarios (py3exiv2, piexif, no exif) with hopefully some smart utilization of pytest fixture parametrization. Unless you have already thought of something, I would just work on it after this PR in some playground branch and ask you for a review.

To get the CI passing I *think* we have to add the actual exiv2 library to the installed dependencies in `.github/workflows/ci.yml`. There are two usages of `apt-get install`, one for the python3.6/3.7 job, one for the python3.8 job. From a quick [package search](https://packages.ubuntu.com/search?keywords=exiv2&searchon=names&suite=bionic&section=all) I hope just adding `exiv2` fixes the problem, but I always have issues with ubuntu packaging :laughing: Could you either try this out or just add the correct packages if you happen to know what is needed on ubuntu?

vimiv/imutils/exif.py
```diff
@@ -10,139 +10,361 @@
...
+                    elif keytype in (
+                        piexif.TYPES.Ascii,
+                        piexif.TYPES.Undefined,
+                    ):  # byte encoded
+                        exif[keyname] = (keyname, val.decode())
+                    elif keytype in (
+                        piexif.TYPES.Rational,
+                        piexif.TYPES.SRational,
+                    ):  # (int, int) <=> numerator, denominator
+                        exif[keyname] = (keyname, f"{val[0]}/{val[1]}")
```
karlch, line 126: Is there any reason why we switched to a dictionary of the form `key -> keyname, value` later in the py3exiv2 version? I don't see us using the actual key anywhere in the formatted metadata, so the initial `keyname -> value` mapping would suffice, no? This would simplify the dict as well as fix the type annotation :smile:

vimiv/imutils/exif.py
```diff
@@ -10,139 +10,361 @@
...
+
+        except (piexif.InvalidImageDataError, KeyError):
+            return None
```
karlch, line 129: I think we should just return an empty dict here, i.e. `return {}`, as we do not deal with `None` explicitly and an empty dict is falsy in itself.

vimiv/imutils/exif.py
```diff
@@ -10,139 +10,361 @@
...
+                    key_name = self._metadata[key].name
+
+                    try:
+                        key_value = self._metadata[key].human_value
+
+                    # Not all metadata(iptc) provide human_value, take raw_value instead
+                    except AttributeError:
+                        key_value = self._metadata[key].raw_value
+
+                    exif[key] = (key_name, key_value)
```
karlch, line 248: Here is where you first introduce the new maping type, this comment is just for reference and possibly adding the explanation here if it has to do with the py3exiv2 implementation.

vimiv/imutils/exif.py
jcjgraf: Thank you, I hope you have had a merry Christmas too! :gift: 

I have added this method in a89d6fdba3b844b36457928852cce5ec13eda2ab

vimiv/imutils/exif.py
jcjgraf: Yes, there is a reason why I index the dict this way. `keyname` is not unique among the metadata (e.g. there is `Iptc.Application2.Copyright` and `Exif.Image.Copyright`, so `Copyright` is not suitable as a key).

What we could instead of using `key -> keyname, value`, we could use a `key -> value` dict and extract the `key_name` from the `key` in `vimiv/metadatawidget`. This should work because `key` has the format `familyKey/groupKey/keyName`.

vimiv/imutils/exif.py
jcjgraf: Thanks for pointing this out. I have fixed it in 7ef003c7dd9f43f36816598bacf6b0fe643ed368

vimiv/imutils/exif.py
karlch: Ah, yes, this makes a lot of sense. Thanks for the explanation!

In this case I would add a new type for the dictionary at the top of the module, i.e.
```
_logger = log.module_logger(__name__)

ExifDictT = Dict[Any, Tuple[str, str]]
```
and use this for the return value type annotation of `get_formatted_exif` instead of `ExifDictT`. For completeness we could also ensure that the value is always of type string. This is currently not the case for integer and float in the piexif case and possibly not for the `raw_value` in the py3exiv2 case, this I am not sure about. I suppose a simple `str(...)` should suffice.

vimiv/imutils/exif.py
karlch: Awesome, thank you!

tests/unit/imutils/test_exif.py
jcjgraf: I have tried to get this to works yesterday for quite some time. All required packages should get installed. The installation of `py3exiv2` fails with `/usr/bin/ld: cannot find -lboost_python38`. According to the internet (e.g. [here](https://stackoverflow.com/questions/59437794/usr-bin-ld-cannot-find-lboost-python-py34) ) a simple symlink should fix it. But I have not had any success yet.

vimiv/imutils/exif.py
jcjgraf: Thanks for pointing this out concerning the strings. I have fixed that as well as the dict type in e2ebcdf32eec9a14763e555a212c3bdccfd618c6

tests/unit/imutils/test_exif.py
karlch: Thanks for taking multiple attempts at fixing this, it seems like this is much trickier than I had hoped. Not a fan of ubuntu packaging in general, but this in combination with doing it on CI seems horrible :worried:

In case you don't find the magic solution by then, I will start a test project to play around with starting on the 2nd of January, I still have some other stuff to do (including celebrating :laughing:) until then. Would you mind if I merge this into a non-master branch first when and if I find some solution to this havoc so I can continue testing before merging into master?

jcjgraf: I am sorry that I have not manages to get it running yet :blush: I have tried everything I could think of as well as everything I found online, but without any success. And to be honest I am  also not sure what else to try.

The compilation step `g++ -pthread -shared -Wl,--rpath=/opt/hostedtoolcache/Python/3.8.6/x64/lib -Wl,--rpath=/opt/hostedtoolcache/Python/3.8.6/x64/lib build/temp.linux-x86_64-3.8/src/exiv2wrapper.o build/temp.linux-x86_64-3.8/src/exiv2wrapper_python.o -L/opt/hostedtoolcache/Python/3.8.6/x64/lib -lboost_python38 -lexiv2 -o build/lib.linux-x86_64-3.8/libexiv2python.cpython-38-x86_64-linux-gnu.so` fails with `/usr/bin/ld: cannot find -lboost_python38`.

I am not at all experienced with C++ compiling (only little C compiling). So, for someone more versed the problem may be obvious. Here things I have tried, thoughts and open questions:
- All required packages should be installed (acording to the [official site](https://pypi.org/project/py3exiv2/) as well as this [SO post](https://stackoverflow.com/a/50381827/5464989))
- Are the packages somehow installed in a wrong order?
- Are some packages conflicting?
- `libboost_python-py38.so` is not available in the image, only the python 3, 3.6, 2 and 2.7 version. So I have symlinked one for version 3.8.
- I have created symlinks at different locations (I not really sure where the compiler/linker looks for them): `/usr/lib/x86_64-linux-gnu`, `/usr/lib/` as well as `/opt/hostedtoolcache/Python/3.8.6/x86/lib` (which is explicitly added as a library source in the `g++` command).
- After creating the symlinks does the linker have to be notified in some way (about the new libraries)? I have tried running `ldconf` after creating the links.

The cause of the problem seems to be somehow related to the ubuntu:latest image of the github CI. When conducting the same steps on docker running ubuntu:latest, everything works as excepted (even without manually linking any libraries).

Since I do not think that I am able to resolve this, I will cleanup the commits in this branch. Then you can can continue experiementing and testing as you want.

karlch: Awesome, thanks a lot once again for this tremendous amount of work. I have merged this into `additional-exiv2` for now so I can force-push, rebase and so forth without all the noise it would make here :blush:

jcjgraf:
Closing this PR in favious of #311 where work continuous.

PR #313, opened 2021-01-05
Mark Speed in Thumbnail Mode

jcjgraf:
When an image is marked in thumbnail mode, the view is repainted right after each mark. This makes marking of many images (e.g. `:mark *` in thumbnail view painfully slow (~55s for ~1350 images). During that time the GUI is also blocked and unresponsive.

To improve that, we add an additional signal which is broadcast from mark methods in `api._mark` once they are done marking all images. This way clients can prolong the expensive repainting.

This change improves the speed a lot (fraction of a second for ~1350 images).

@karlch if you belief that his is the wrong way to fix this problem and you have got a better idea let me know.

**EDIT**
Argh, this pipeline again :unamused:

### DIFF
--- a/vimiv/api/_mark.py
+++ b/vimiv/api/_mark.py
@@ -27,6 +27,7 @@ class Mark(QObject):
     Signals:
         marked: Emitted with the image path when an image was marked.
         unmarked: Emitted with the image path when an image was unmarked.
+        markdone: Emitted when all image of a given paths list are (un)marked.
 
     Attributes:
         _indicator: Attribute to cache the evaluated mark indicator string.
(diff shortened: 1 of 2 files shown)


karlch:
Makes sense to me, thanks for the PR! I think everything should be fine if you remove the lambda and just connect the method directly, i.e. `markdone.connect(self.repaint)`. Could you please also document the new `markdone` signal in the docstring of the `Mark` class?

jcjgraf:
You were right, the lambda somehow caused the CI to break. I have also added the missing docstring. I hope you like it.

PR #319, opened 2021-01-17
Add :metadata-list-keys command

### LINKED ISSUE #220: Add --list-keys option to :metadata command

karlch:
Introduces a new `:metadata-list-keys` command to display all valid exif keys for the current image.

I decided to implement this as a standalone command instead of an option for `:metadata` as there are two optional arguments for `:metadata-list-keys`:
- `--n-cols` to customize the number of columns used to format the html table
- `--to-term` to print to the terminal instead of displaying the table in case one would like to easily copy / format / pick ...

These would otherwise just clutter the `:metadata` command.

@jcjgraf what do you think? One point where I would like to hear your opinion is that I removed all keys which are plain hex digits, like `Exif.CanonCs.0x0006`, I suppose this makes sense?

fixes #220

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -31,6 +31,8 @@ Added:
   exif values into a human readable format. Thanks a lot
   `@jcjgraf <https://github.com/jcjgraf>`_ for all your hard work, thoughts and comments
   on this topic!
+* New ``:metadata-list-keys`` command to display all valid exif keys for the current
+  image.
 
 Changed:
 ^^^^^^^^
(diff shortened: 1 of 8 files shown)


jcjgraf:
This looks really good to me. I do not have anything to add or criticise.

I was not aware that there are unknown metadate keys, but according to the exiv2 docs there are few. It is probably best to just ignore them, as you are doing.

Thanks for adding this feature!

PR #321, opened 2021-01-20
Update Metadata Default Keys

jcjgraf:
After merging #311 I thought it is a good time to rework the default metadata keys. Now there are a total of 5 sets which contain the following information:
1. General information about the image
2. Exif camera settings
3. GPS information
4. IPTC data
5. Technical details about the image

I am really happy with the capabilities of metadata widget. But in the set 5 I wanted to present information like color space, image resolution, image format, image size etc. But, not all of these information are stored in exif (or some only for uncompressed files). To get these information into the metadata widget, I thought that we may extend the `ExifHandler` to accept some custom keys, which are dealt with by vimiv itself and not my exiv2. Concretely, I have thought about the following keys:
- `vimiv.ImageFormat`: Image extension (JPEG, CR2 etc.)
- `vimiv.XDimension`: Length of image in pixels
- `vimiv.YDimension`: Height of image in pixels
- `vimiv.FileSize`: The filesize in MB

I think these information are sometimes quite crucial. The proposed solution has the advantage that these additional keys are well integrated into the metadata widget. Meaning that for the user it does not make any difference where the data comes from, it is just all presented in the metadata widget.

In addition, I have had the following idea. The raw coordinates presented in 3. are not very meaning full (at least not to me). So I thought it would be awesome to have a little map which show where a certain image is taken. It does not have to be complex, just a map with zoom and tilt functionality and with a red dot indicating where the image was taken. But that should probably rather go into a dedicated plugin :blush: 

Please let me know that you think about the chosen defaults and if would would find it useful too to have some additional metadata keys.

### DIFF
--- a/tests/end2end/features/image/conftest.py
+++ b/tests/end2end/features/image/conftest.py
@@ -24,6 +24,7 @@ def exif_content():
             piexif.ImageIFD.Copyright: b"vimiv-AUTHORS-2020",
         },
         "Exif": {piexif.ExifIFD.ExposureTime: (1, 200)},
+        "GPS": {piexif.GPSIFD.GPSAltitude: (1234, 1)},
     }
 
 
--- a/tests/end2end/features/image/metadata.feature
+++ b/tests/end2end/features/image/metadata.feature
@@ -30,8 +30,8 @@ Feature: Metadata widget displaying image exif information
         When I add exif information
         And I run 3metadata
         Then the metadata widget should be visible
-        And the metadata text should contain 'Copyright'
-        And the metadata text should contain 'vimiv-AUTHORS-2020'
+        And the metadata text should contain 'GPSAltitude'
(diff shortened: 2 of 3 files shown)


karlch:
Awesome, thank you! The grouping makes a lot of sense to me. As I don't use the metadata very often, I am totally fine with whatever you propose. If you agree, I would merge this as-is and keep the other points you describe separate.

Could you open an issue concerning the "vimiv metadata" keys? We could then keep the discussion in one place.

Concerning the fancy GPS map: This certainly sounds really cool, but also rather complex and out of scope for vimiv. Having it in a plugin would be amazing though! In case you would like to tackle this, we can certainly discuss if there is anything needed from the vimiv side and I am always happy to help if there are any questions in general.

jcjgraf:
Feel free to merge this PR, I will open seperate issues for the other things.

PR #322, opened 2021-01-20
More vim movements

### LINKED ISSUE #295: More vim movement shortcuts in thumbnail/library mode
As recommended by @jcjgraf in #294 we could implement more of vim's movement shortcuts.

**thumbnail mode:**
- `^` and `$` to move to the beginning/end of the current row (I would prefer not to touch `0` as it is interpreted as a digit)
- `<ctrl>u`, `<ctrl>d`,  `<ctrl>b`,  `<ctrl>f` to scroll down/up (half) a page

**library mode:**
- `<ctrl>u`, `<ctrl>d`,  `<ctrl>b`,  `<ctrl>f` to scroll down/up (half) a page

karlch:
Implements additional vim movement commands as proposed in #295.

As you came up with the idea, would you like to give this a test and check if this is what you had in mind before I merge @jcjgraf?

fixes #295.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -6,6 +6,18 @@ All notable changes to vimiv are documented in this file.
 v0.9.0 (unreleased)
 -------------------
 
+Added:
+^^^^^^
+
+* More vim movement commands with corresponding keybindings, namely:
+
+  * ``:scroll page-up/page-down/half-page-up/half-page-down`` mapped to ``<ctrl>b``,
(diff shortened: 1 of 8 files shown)


jcjgraf:
This is awesome. The shortcuts do exactly what I would except them to do. Except the `^` needs getting used to since I always use the `0` for that purpose. But I except that it would be a bit troublesome to bind the zero key.

Thanks for you effort!

Issue #324, opened 2021-01-21
Vimiv Metadata Keys


jcjgraf:
As briefly mentioned in #321, it would be useful to to be able to display some information about the image file in the metadata widget, which are not part of exif (or only part for some formats). Concretely, it would be useful to be able to get the imageformat, X and Y dimension in pixels and the file size.

To integrate this information into the metadata widget, probably the easiest option would be to add some special keys, which are e.g. prefixed by `Vimiv.<key>`. In `imutils.exif.ExifHandler` we would have to filter these options before submitting the keys to exiv2 and handle them separately.

The advantage of this approach is that these additional information is well integrated into the metadata widget, meaning that no extra command or something is required to display them.

Open questions:
- What keys would be useful to have? I have so far come up with:
    - `Vimiv.ImageFormat`: Image extension (JPEG, CR2 etc.)
    - `Vimiv.XDimension`: Length of image in pixels
    - `Vimiv.YDimension`: Height of image in pixels
    - `Vimiv.FileSize`: The file size in MB
- Should `imutils.exif.ExifHandler` be renamed to `imutils.metadata.MetadataHandler` (or something similar) since the handler deals with exif, iptc and general metadata.

What do you think @karlch? Do you have a better idea for an implementation?

karlch:
I like the idea of having additional "internal" keys, but I am not sure I like mixing this with the `ExifHandler` directly. The code is already quite involved there. I am fine with renaming the handler though, as you say also iptc is supported.

Personally, I would prefer to have all of this split into different chunks, maybe something like:
- The exif / metadata handler that we have currently to deal with regular keys
- Some function / class / module to deal with vimiv keys
- Some class that combines the two and delegates the work accordingly

Careful thinking would be especially important if the vimiv keys require accessing the current pixmap to retrieve the pixel size as I would really like to keep exif and qt pixmaps as separate as possible. The more I think about this the more afraid I am of possible code coupling, but I may be overthinking and should probably wait for some actual coding :laughing: 

What do you think? Do you want to take a shot at this?

Issue #323, opened 2021-01-20
Add Functionality to API


jcjgraf:
While working with the api for some plugins I noticed some missing functionality:

 - `api.mark`: Additionally, it should expose:
    - `api.mark.is_marked(name: str)` to check if a image is marked. Currently this required accessing the private `api.mark._marked` list.
    - `api.mark.mark2(name: str)` which marks the image no matter is it is already marked or not. Currently this required checking if the image is already marked and use `api.mark.mark` only when it is not.
    - `api.mark.unmark(name: str)` which unmarks the image no matter if it is marked or not. This and the previous functionality could also be added to `api.mark.mark` using certain flags.
    - Since when marking multiple images, the `api.mark._markdone` signal has to be emitted manually it may be a good idea to provide the above mentioned functionality also for a list of paths. Currently one has to add/remove all images to/from `api.mark._marked` and then manually emit the signal.

- `imutils.exif.ExifHandler`: We should add something like `api.get_exifhander(path: str) -> ExifHandler` which returns the exif handler for a certain path. Currently this is done `imutils.exif.ExifHandler(path)`.

Besides that, one has to use `utils.log.module_logger` to get a logger, and to access the vimiv colors one has to use `config.styles`. For the status bar module I have also come across using `utils.wrap_style_span`.

I am not sure where we have to draw the line between what to expose to the API and what is acceptable to access directly. :blush:  I think my first two points would be acceptable to add. But let me know what you think.

karlch:
I fully agree with your first two points. The exif handler function should be very straightforward as you point out. The different utility methods in `mark` may need some thinking and possibly refactoring so they all have useful and intuitive names.

Everything in `utils` is certainly borderline. It does not really make sense to expose the little functions like `wrap_style_span` via a dedicated api, but of course they can be useful for plugins as well. I guess "use at own risk" should be fine for those, the chances of them being changed in a backwards-incompatible way is rather small anyway.

Getting a logger directly from utils should be safe, I don't see that infrastructure changing (anytime soon). Do you think it makes sense to have a dedicated `api.utils` which imports parts which are considered safe from `utils` or is documentation enough?

I would not expose the styles in their current state as I am not 100 % happy with the string-based usage all over the place, so there are quite some chances that this could change at some point :laughing:

jcjgraf:
I do not really have a strong opinion about the `utils`. I rather wanted to get your view concerning non-api functionality from plugins. I am fine with whatever you think is right.

[...]

PR #325, opened 2021-01-22
Internal metadata keys

jcjgraf:
This PR adds internal metadata keys to vimiv, as outlined in #324. Unlike @karlch has suggested, I have put everything into the preexisting `ExifHandler`. I had the feeling that the proposed idea of having separate classes for the internal and external keys is slightly overkill for the few internal keys we are actually adding.

The basic idea is the following: Clients do not longer call the ExifHandlers child-specific `get_formatted_exif` but `get_formatted_metadata`. This method is not overwritten by any child. It iterates over all `desired_keys` and handles all internal keys directly. For the exif keys, the child-specific `_fetch_external_key` is called which extracts the exif using the appropriate library. `get_formatted_metadata` combines all the extracted data and returns it to the client.

To get the image dimension I use the `QImageReader` class instead of accessing the pixmap somehow. This has the advantage that it is fully decoupled qt pixmap of vimiv. This reader class is actually meant for accessing image properties without rendering a fullblown pixmap.

In my view, the new functionality could be added the `ExifHandler` in a rather clean manor and the complexity of the class is still reasonable. 

Nevertheless, this PR is just a proposition and in the current state merely a proof-of-concept. If you think that it is still better to split the functionality I will redo this PR.

If we use this PR as a basis, there are the following TODOs:
- [ ] Rename class, methods and docstring to metadata instead of exif
- [ ] Add/update docstrings of new/changed methods.
- [ ] How should we handle the case when no exiv library is installed? Just present these 4 lonely keys? :laughing: 
- [ ] How to represent the keys internally? Currently, all keys are stored in a list, but in the `get_formatted_metadata` I still redefine them multiple times. Would something like a dict cleanup the mess?
- [ ] Are there any other useful internal keys?
- [ ] Update docs.
- [ ] Better way to create the generator in `get_keys`?


karlch:
That was fast, awesome, thanks for the PR! I really like using `QImageReader` for this, that makes a lot of sense.

I still feel like an additional class for the internal keys could make sense, especially if it inherited from `dict` so we could simplify the `get_formatted_metadata` to something like:
```
    def get_formatted_metadata(self, desired_keys: Sequence[str]
    ) -> Dict[Any, Tuple[str, str]]:
        metadata = dict()

        for base_key in desired_keys:
            try:
                key, key_name, key_value = self._fetch_key(base_key)
                metadata[key] = key_name, key_value
            except (KeyError, TypeError):
                _logger.debug("Invalid key '%s'", base_key)

        return metadata

    def _fetch_key(self, key: str):
        if key.lower().startswith("vimiv"):
            return self._internal_handler[key]
        return self._fetch_external_key(key)
```
If it were in it's own class, the class could also lazy-initialize the `QImageReader` to avoid constructing it multiple times, i.e. something like
```
class _InternalKeyHandler(dict):

    def __init__(self, path: str):
        super().__init__(
            # Define all the keys here with the corresponding actions
        )
        self._path = path
        self._reader: Optional[QImageReader] = None

    @property
    def reader(self) -> QImageReader:
        if self._reader is None:
            self._reader = QImageReader(self._path)
        return self._reader
```
We would then override `__getitem__` in the internal key handler to:
- Find the corresponding action case-insensitvely
- Call the action to get the corresponding value
- Return in the same format as `_fetch_external_key`

This is getting rather long, but I will try to add my thoughts on some of the other open issues:
- I am not sure if we should expose the 4 lonely keys. If so, we may need to do some refactoring in the metadata widget, and certainly remove the `has_exif_support` part.
- If we go for the internal key handler class, this class would store and encapsulate all the keys.
- I really like the way you chain the two parts into the `get_keys` generator. Possibly this logic could only be done once in the base class, so we don't have to repeat the code concerning chaining / getting the internal keys in the children. Very similar to how you split `get_formatted_metadata` and `_fetch_external_key`.

jcjgraf:
Thank you for your comments and suggestions. I will look into it and see what I can do. But be patient, I am really busy the new two weeks an I am unsure when I can look into this.

karlch:
Awesome, thank you, and as always: no hurry! :blush:

jcjgraf:
I have split the `ExifHander` class into three separate ones; `MetadataHandler` is the main class and it is used by clients to interact with metadata. `InternalKeyHandler` deals with the internal keys and it build upon the framework you have given me. `ExternalKeyHandler{Base, Piexif, Exiv}` is the old `ExifHander{Base, Piexif, Exiv}`.

There is still lots of cleanup etc. to do. But before I do that I would be grateful if you could have a glimpse at the code and comment on the general structure. :blush:

karlch:
So you did end up adding another class which combines the two handlers. Personally, I like this a lot, we keep the fetching and combining logic separated. Only downside is the need to more or less redefine the functions for `copy_metadata` and `get_date_time`. How about also exposing the `ExternalKeyHandler` so we can call `copy_metadata` and `get_date_time` directly in the respective parts of the code, and only use `MetadataHandler` to get the metadata keys and information?

I will add a few inline comments to clarify some of my ideas :blush:

vimiv/imutils/exif.py
```diff
@@ -261,13 +293,95 @@ def copy_exif(self, dest: str, reset_orientation: bool = True) -> None:
...
+    ) -> Dict[Any, Tuple[str, str]]:
+        metadata = dict()
+
+        for base_key in desired_keys:
+            base_key = base_key.strip()
+
+            try:
+                key, key_name, key_value = self._fetch_key(base_key)
+                metadata[key] = key_name, key_value
+            except (KeyError, TypeError):
```
karlch, line 362: How about also catching the `UnsupportedExifOperation` here for completeness, instead of handling it in `_fetch_keys` directly? We will probably never use the class without exif support, but safety first, you are right :blush:

vimiv/imutils/exif.py
```diff
@@ -261,13 +293,95 @@ def copy_exif(self, dest: str, reset_orientation: bool = True) -> None:
...
+
+        return metadata
+
+    def get_keys(self) -> Iterable[str]:
+        """Retrieve the name of all exif keys available."""
+        try:
+            return itertools.chain(
+                self.internal_handler.get_keys(), self.external_handler.get_keys()
+            )
+        except UnsupportedExifOperation:
```
karlch, line 373: I guess we would have to return an empty list or so. But I am also fine with letting the exception propagate, like this the caller can directly decide what to do when there is no exif support.

vimiv/imutils/exif.py
```diff
@@ -261,13 +293,95 @@ def copy_exif(self, dest: str, reset_orientation: bool = True) -> None:
...
+    def copy_metadata(self, dest: str, reset_orientation: bool = True) -> None:
+        """Copy exif information from current image to dest.
+
+        Args:
+            dest: Path to write the exif information to.
+            reset_orientation: If true, reset the exif orientation tag to normal.
+        """
+        try:
+            self.external_handler.copy(dest, reset_orientation)
+        except UnsupportedExifOperation:
```
karlch, line 339: If we decide to keep these methods in the metadata handler, I would certainly let the exception propagate. But not having them here at all, seems even nicer to me :blush:

jcjgraf:
Yeah, as you initially suggested, three classes allow to split these different modules in a clean manor.
I am against exposing the `ExternalKeyHandler`. I think the clients just want the metadata and they should not have to bother if the data is coming from the `ExternalKeyHandler` or from the `MetadataHandler`.

karlch:
Oh, sorry, the wording "expose" was probably rather bad. I fully agree that no-one should have to bother where metadata comes from. What I meant was to make the `_ExternalKeyHandler` class public for usage within vimiv so we don't have to wrap the two functions `copy_metadata` and `get_date_time`. We currently only use these internally in `imutils._file_handler` / `imutils.filelist` and I don't really see the point in having them in api-related functionality after exposing the metadata handler to the api as proposed in #323.

In this case I actually wonder if it also makes sense to make `MetadataHandler._fetch_key` public or even make it `__getitem__` so, once exposed via the api, it is easy for clients to retrieve single keys.

jcjgraf:
Ah, this makes much more sense. I thought you are talking about exposing it to the API :upside_down_face: 
Thanks for these suggestions and ideas! I will look into it in the following week.

vimiv/imutils/exif.py
jcjgraf: I have removed them in d3281b46867badf1db503c8be7375504cbfc9b59

vimiv/imutils/exif.py
jcjgraf: I think it makes more sense to just propagate the exception. I have done that in 31cddcb942705a911d07823a6ab013bc41fddfb3. Additionallty, I have added a notice to the docstrings that clients must fetch the UnsuportedExifOperation exception.

vimiv/imutils/exif.py
jcjgraf: As with the `get_keys` methods, I think it makes sense to propagate the excepion here too since this method can be called by cliets via the API anytime.

vimiv/version.py
```diff
@@ -42,8 +42,8 @@ def info() -> str:
         f"Qt: {QT_VERSION_STR}\n"
         f"PyQt: {PYQT_VERSION_STR}\n\n"
         f"Svg Support: {bool(QtSvg)}\n"
-        f"Pyexiv2: {exif.pyexiv2.__version__ if exif.pyexiv2 is not None else None}\n"
-        f"Piexif: {exif.piexif.VERSION if exif.piexif is not None else None}"
+        f"Pyexiv2: {metadata.pyexiv2.__version__ if metadata.pyexiv2 is not None else None}\n"
```
jcjgraf, line 45: This line causes the linter to fail (too long). Whats is the best was to fit it? To use a variable ro the piexiv version?

jcjgraf:
Commit 71ac242a6f8409673d406808ffad699ce6d06d5b has somehow broken the `test_switch_metadata_information_upon_new_image` testcase. Do you have an idea what went wrong (I am sorry, I am not very familiar with unit tests).

How should we unit test this module? If I am not mistaken the metadata functionality is only tested via the metadata widget. Probably it would make sense to test these handlerclasses separately.

vimiv/version.py
karlch: I would just break it like this:
```python
        f"Svg Support: {bool(QtSvg)}\n"
        "Pyexiv2: "
        f"{metadata.pyexiv2.__version__ if metadata.pyexiv2 is not None else None}\n"
        f"Piexif: {metadata.piexif.VERSION if metadata.piexif is not None else None}"
```

vimiv/gui/metadatawidget.py
```diff
@@ -121,7 +121,12 @@ def metadata_list_keys(self, n_cols: int = 3, to_term: bool = False):
                 * ``--to-term``: Print the keys to the terminal instead.
             """
 
-            keys = sorted(set(self.handler.get_keys()))
+            try:
+                keys = sorted(set(self.handler.get_keys()))
+                _logger.debug("Successfully got keys")
+            except metadata.UnsupportedMetadataOperation:
```
karlch, line 127: I would just not catch the exception here. The code would break anyway if the keys are not defined, but this should never happen as you say.

vimiv/gui/metadatawidget.py
```diff
@@ -154,15 +160,21 @@ def _update_text(self):
...
                 e.strip() for e in api.settings.metadata.current_keyset.value.split(",")
             ]
             _logger.debug(f"Read metadata.current_keys {keys}")
-            formatted_exif = self.handler.get_formatted_exif(keys)
-            if formatted_exif:
-                self.setText(utils.format_html_table(formatted_exif.values()))
+            try:
+                data = self.handler.fetch_keys(keys)
+                _logger.debug("Fetched metadata")
+            except data.UnsupportedMetadataOperation:
```
karlch, line 172: Same thing here.

vimiv/imutils/metadata.py
```diff
@@ -0,0 +1,418 @@
...
+    def __getitem__(self, key: str) -> Tuple[str, str, str]:
+        """Intrypoint to extract the key of the image at _path.
+
+        Args:
+            key: internal key to fetch.
+        """
+        try:
+            key, func = super().get(key.lower())
+            return (key, key, func())
+        except KeyError:
```
karlch, line 79: If we don't catch the `KeyError` here, the metadata widget can deal with it accordingly and we don't end up with empty entries in the dictionary of the form `"": ("", "")`. This is actually what breaks the tests, it only adds the `no metadata information found` line if the dictionary is completely empty, but false keys are added with empty strings like this.

vimiv/imutils/metadata.py
```diff
@@ -0,0 +1,418 @@
...
+                    if keytype in (
+                        piexif.TYPES.Rational,
+                        piexif.TYPES.SRational,
+                    ):  # (int, int) <=> numerator, denominator
+                        return (keyname, keyname, f"{val[0]}/{val[1]}")
+
+        except (piexif.InvalidImageDataError, KeyError):
+            return ("", "", "")
+
+        return ("", "", "")
```
karlch, line 195: The same thing happens here. I would rewrite this like this:
```python
with contextlib.suppress(piexif.InvalidImageDateError):
    # All the key handling here
    
raise KeyError(f"Key '{base_key}' not found")
```
so the metadata widget can ignore the key by catching the exception.

vimiv/imutils/metadata.py
```diff
@@ -0,0 +1,418 @@
...
+                        key_value = ", ".join(value)
+                    else:
+                        key_value = value
+
+                return (key, key_name, key_value)
+
+            except KeyError:
+                _logger.debug("Key %s is invalid for the current image", key)
+
+        return ("", "", "")
```
karlch, line 295: Once again, re-raising the key error would save the day :blush:

vimiv/imutils/metadata.py
```diff
@@ -0,0 +1,418 @@
...
+        return self.reader.size().height()
+
+    def __getitem__(self, key: str) -> Tuple[str, str, str]:
+        """Intrypoint to extract the key of the image at _path.
+
+        Args:
+            key: internal key to fetch.
+        """
+        try:
+            key, func = super().get(key.lower())
```
karlch, line 77: I would use `super().__getitem__(key.lower())` here, as `get` gives you `None` instead of raising an exception.

karlch (changes requested):
This is looking great! I left a few comments that should hopefully fix the tests as well as make the complete CI pass. Have a nice week-end! :blush:

karlch:
Concerning the unit-tests: I agree that we should probably test them all in a nice and isolated fashion in `tests/unit/test_metadata.py`. As we no longer have the setting defining which metadata keys to get, this should be possible quite nicely by first setting up an image with known exif content, and then testing the individual parts `fetch_key` and `get_keys` once on the individual handlers and then maybe even on the combined one.

If you like, you can take a shot at this and ask any questions. Otherwise I am happy to do this after merging.

jcjgraf:
Thank you for your suggestions, hits and comments. I will go though it in the following week(s) and also give it a try to create some appropriate unit test. Though be warned, I have to not that before :smile:

vimiv/gui/metadatawidget.py
jcjgraf: Fixed in 16ec200

vimiv/gui/metadatawidget.py
jcjgraf: Fixed in 16ec200

vimiv/imutils/metadata.py
jcjgraf: Makes sense. Thank you for spotting that. Though, the metadata widget used `metadata/fetch_keys` which handles `KeyError`. Onl `metadata/fetch_key` throws `KeyError`. So, there is not need to chany anything in metadata widget.

Fixed in 16ec200.

vimiv/imutils/metadata.py
jcjgraf: Fixed in 16ec200

vimiv/imutils/metadata.py
jcjgraf: Fixed in 16ec200.

vimiv/imutils/metadata.py
jcjgraf: Well spot! I am not of fluent enough in Python to know such details :upside_down_face: 

Fixed in 16ec200.

karlch:
Awesome work, this is taking getting in good shape! :blush:

I noticed that the `noexif` and the `piexif` environments are failing, both are trivial fixes:
- `noexif`: We apply a custom skip marker to the tests in the `test_exif.py` unit test in `tests/conftest.py`. As the test was renamed, the marker has to be renamed as well. This is in the beggining of the `apply_exif_markers` function.
- `piexif`: I had a typo in the code snippet I sent you. It should be `contextlib.suppress(piexif.InvalidImageDataError)` not `contextlib.suppress(piexif.InvalidImageDateError)`. Sorry :smile: 

Finally, you somehow managed to wreck havoc with the git history :laughing: I guess a simple rebase against the current master here should fix this.

karlch:
Would you like some more time to work on this or is it time to take another look and hopefully provide some help? :blush:

jcjgraf:
I am sorry that I have neglected this PR a little in the last few weeks :relaxed: 

It would actually be great if you could give me some hints any tips. I am kind off stuck with two things:
- Concerning the test setup and preparation of a dummy test image. `Metadatahandler`requires the path to an image (and not an instance if QImage or something). Therefore, in `imutils/test_metadata.py`  have added the fixture `dummy_image()` which creates a image using PIL and saves it to the disk.
    - Is this actually the right approach or is there some cleaner method?
    - How to remove the image after test? Just create a fixture wich runs something like `os.system("rm ./image.jpg")` after the test?
   
- I have started with a first test `test_MetadataHandler_fetch_key()` which should test the `MetadataHandler.fetch_key()`. In this test, firstly I get a dictionary of `pyexiv2.exif.ExifTag` instances (returned by `metadata_content()`). Then I iterate over all elements of this dictionary in order to check if my prepared image contains all these metadata keys/values. Weirdly, accessing properties of these `pyexiv2.exif.ExifTag` instances while iterating over the dictionary causes a segmentation fault.
    - Now, I am unsure if I do make some obvious mistake or if this is a non-trivial problem (e.g. a problem with pyexiv2).
    - Currently, the test just fails but does not report segfault. So see the segfault change the test to e.g.
```
def test_MetadataHandler_fetch_key(get_MetadataHandler, metadata_content):
    for key, value in metadata_content.items():
        print(value.human_value)
    assert True
```

After solving the first two points, a follow up question is the following. Let's say I have defined the test `test_A`, test_B`, ... `test_Z`. How do I run these test on piexif, pyexiv2 and noexif? I guess it is somehow relate to monkeypatches but I have not yet figured out how that all words.  

I hope explaining me how to write these tests does not take more time if you would just write them yourself.

karlch:
No worries at all, I have neglected work on vimiv completely over the past months ...

This took some debugging, and it indeed required writing the code, so I will just dump the snippets I came up with and some hopefully useful information for future PRs :blush: 
```python
@pytest.fixture()
def dummy_image(qapp, tmp_path):
    filename = str(tmp_path / "image.jpg")  # We use pytest's tmp_path fixture to create the image in a tmp directory that is cleaned for us
    QPixmap(300, 300).save(filename)  # Let's use PyQt directly instead of another external module, requires the "qapp" fixture to create a default QApplication
    return filename


@pytest.fixture
def metadata_handler(add_metadata_information, dummy_image, metadata_content):
    assert pyexiv2 is not None, "pyexiv2 required to add metadata information"
    add_metadata_information(dummy_image, metadata_content)
    return MetadataHandler(dummy_image)


@pytest.fixture
def metadata_handler_piexif(dummy_image, metadata_handler):
    # This is certainly not beautiful, but I could not come up with a consistent solution as we need pyexiv2
    # to add the metadata information and the external handler is defined upon import level
    metadata_handler._ext_handler = metadata._ExternalKeyHandlerPiexif(dummy_image)
    return metadata_handler


def test_metadatahandler_fetch_key(metadata_handler, metadata_content):
    for key, value in metadata_content.items():
        fetched_key, _, fetched_value = metadata_handler.fetch_key(key)
        assert fetched_key == key
        try:
            assert fetched_value == value.human_value
        except AttributeError:
            assert fetched_value == value.raw_value


def test_metadatahandler_fetch_key_piexif(metadata_handler_piexif, metadata_content):
    # I also ended up writing an additional test for piexif as
    # 1) the keys are in their short form and thus must be compared differently
    # 2) there is no support for the human value also making this comparison different
    for key, value in metadata_content.items():
        fetched_key, _, fetched_value = metadata_handler_piexif.fetch_key(key)
        short_key = key.rpartition(".")[-1]
        assert fetched_key == short_key
        assert fetched_value == value.raw_value
```
and I had to modify the `add_metadata_information` fixture as follows:
```python
        for tag in content.values():
            _metadata[tag.key] = tag.value
```

It seems like pyexiv2 invalidates the python `ExifTag` object after writing and reading. Here is a short script that can reproduce the segfault without any of the vimiv-related stuff:
```python
import pyexiv2
from PyQt5.QtGui import QImage


tag = pyexiv2.ExifTag("Exif.Image.Copyright", "vimiv-AUTHORS-2021")

path = "image.jpg"
image = QImage(300, 300, QImage.Format_ARGB32)
image.save(path)

handler = pyexiv2.ImageMetadata(path)
handler.read()
# Uncomment for segfault
# handler[tag.key] = tag
handler[tag.key] = tag.value

handler.write()

handler = pyexiv2.ImageMetadata(path)
handler.read()

print(tag.human_value)
```

Concerning the piexif part:
In the quick snippet I ended up defining an additional fixture to force using piexif and not pyexiv2 as the test needed to be adjusted anyway. Once there are tests that run as-is with both handlers, we should be able to make us of the parametrized `external_handler` fixture in combination with another `metadata_handler_both(external_handler)` fixture that updates the external handler much like I did for the piexif one. When this is then used in any tests, the tests are run with both (all) parameters of the fixture parametrization.

Let me know if there are any questions and happy hacking :blush:

TornaxO7:
This pull request would probably solve this issue (it pops up, if I open vimiv within vifm):
image

Just mentioning this for "motivation" ;)

jcjgraf:
Thanks @TornaxO7 for using vimiv and reporting this incidence! It is a bit unfortunate that vifm reports this an error but it is actually by purpose. You are missing the optional dependency which is required for metadata support. See here for the link to the required library: https://karlch.github.io/vimiv-qt/documentation/exif.html

@karlch Maybe we should improve this log by indicating how to enable metadata support to prevent further confusion, shouldn't we?

karlch:
Welcome to vimiv and thanks from me as well @TornaxO7.

Yes, I fully agree with everything you said @jcjgraf! Something simple like
```python
    _logger.warning(
        "There is no exif support and therefore:\n"
        "1. Exif data is lost when writing images to disk.\n"
        "2. The `:metadata` command and associated `i` keybinding is not available.\n"
        "3. The {exif-date-time} statusbar module is not available.\n\n"
        "Please install pyexiv2 or piexif to silence this warning.\n"
        "For more information see\n"
        "https://karlch.github.io/vimiv-qt/documentation/exif.html.\n"
    )
```
could suffice IMHO. If you agree with this and the wording, I can just commit this snippet to master.

jcjgraf:
@karlch I had something like that in mind too. Feel free to commit it

TornaxO7:
Installing `python-piexif` solved my issue! The error message doesn't appear in vifm anymore :) Thank you for the hint and for this program guys!

jcjgraf:
Thank you @karlch very much for debugging my segfault and helping with building the framework for testing these module.

I have come up with a series of test which should test all functionality of `MetadataHandler` as well as the internal and external handler. I have not written tests for subfunctions which are called by functions which I have written tests for. E.g. I do not explicitely test `_InternalKeyHandler._get_xdimension()` since this is already covered by the test of `MetadataHandler.fetch_key("Vimiv.XDimentsion")`. I hope this is alright.

A constructive feedback and suggestions for improvements are very welcome since this is the most extensive testset I have written so far :laughing: Especially, if something is not covered by any test.

jcjgraf:
This PR is finally ready for an in-depth review. I hope my commits are not too messy. I did not dear to clean it up more. I fear that I would break something :smile:

I am not sure what happend with the CI pipeline. Somehow it has gone :confused: Probably due to the conflicts with the master.

Commit 3050b22 is "cherry-pick" from #371 since it turned out to be a little bit too painfull to rebase this branch with master. Meaning, the conflicts with `docs/documentation/exif.rst` and `vimiv/imutils/exif.py` can be ignored (overwrite changes to master). While changes to `docs/changelog.rst` require manuel merging.

tests/unit/imutils/test_metadata.py
```diff
@@ -0,0 +1,335 @@
...
+def external_keyhandler_piexif(add_metadata_information, dummy_image, external_content):
+    assert pyexiv2 is not None, "pyexiv2 required to add metadata information"
+    image = dummy_image()
+    add_metadata_information(image, external_content)
+    return metadata._ExternalKeyHandlerPiexif(image)
+
+
+def test_external_keyhandler_get_date_time(
+    external_keyhandler, external_keyhandler_piexif, dummy_image
+):
```
karlch, line 319: This is a really funky way to do this as pytest provides parametrization precisely for this purpose :blush:

As an example, the (unused?) fixture at the top called `external_handler` will make tests run twice, once with `metadata.ExternalKeyHandler` as value for `external_handler` and once with `metadata._ExternalKeyHandlerPiexif` as the argument. This fixture could easily be modified to actually create the handler with some image as path, but it might be nice to do this within this test, so the structure would be something along:
- add metadata to dummy image if any
- create handler with path to dummy image
- compare to expected result

Both the metadata and the expected results could also be parameters, now not of the fixture but of the test itself. The structure would end up looking somewhat like this (dummy code):
```python
@pytest.mark.parametrize("metadata, expected_date", [({"name": ExifTag("name", "value"}, "value"), ({}, "")]
def test_external_keyhandler_copy_metadata(external_handler, dummy_image, add_metadata_information, metadata, expected_date):
    src = dummy_image("image.jpg")
    add_metadata_information(src, metadata)
    handler = external_handler(src)
    assert handler.get_date_time() == expected_date
```

While this doesn't look much better than the four assert statements in your implementation, it does have a few advantages:

- Each assert is run in its own test, i.e. we test pyexiv2 and piexif as well as content / no content separately
- We can very easily add more combinations

This kind of logic applies to some more parts of the tests you wrote, e.g. the copy_metadata test could then also be parametrized instead of having a for loop. Hope this makes sense :blush:

karlch (commented):
Thanks again, this was a lot of great work!

The tests test what they should for the most part, but I feel like the structure could be improved in general. There are an awful lot of fixtures, maybe this could also be broken down with some more parametrization? If there is anything I can help with, just let me know!

None of this is super-critical, so feel free to tell me in case you don't want to go into the havoc of refactoring the tests, I can also take a look at this some other time. On the other hand, if you feel like it could be a fun way to play around with pytest, just go ahead and hack away :blush:

docs/documentation/metadata.rst
```diff
@@ -0,0 +1,91 @@
...
+
+   ======================= ============== ==================== =====================================================================
+   PROPERTY                `piexif`_      `pyexiv2`_           Note
+   ======================= ============== ==================== =====================================================================
+   Exif Support            True           True                 pyexiv2 can extract way more data for the same image
+   ICMP Support            False          True
+   XMP Suppport            False          True
+   Output Formatting       False          True                 e.g. ``FNumber: 63/10`` vs ``FNumber: F6.3``
+   Supported File Types    JPEG, TIFF     All common types
+   Ease of installation    Simple         More complicated     pyexiv2 requires some dependancies including the C++ library `exiv2`_
```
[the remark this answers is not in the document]
jcjgraf: rebased and fixed in 93ee9dde4734c608394e201e46bc14b9bbf9126c

docs/documentation/metadata.rst
```diff
@@ -0,0 +1,91 @@
...
+
+You can configure the information displayed by the ``:metadata`` command by adding your
+own key sets to the ``METADATA`` section in your configfile like this::
+
+    keys2 = Override,Second,Set
+    keys4 = New,Fourth,Set
+
+where the values must be a comma-separated list of valid metadata keys.
+
+In case you are using `pyexiv2`_ you can find a overview of valid Exif, IPTC and XMP
```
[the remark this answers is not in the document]
jcjgraf: rebased and fixed in 93ee9dde4734c608394e201e46bc14b9bbf9126c

vimiv/imutils/metadata.py
```diff
@@ -0,0 +1,440 @@
...
+
+        Throws:
+            KeyError
+        """
+        key, func = super().__getitem__(key.lower())
+        return (key, key, func())
+
+    def get_keys(self) -> Iterable[str]:
+        """Returns a sequence of all available metadata keys."""
+        return (key for key, _ in super().values())
```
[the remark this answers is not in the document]
jcjgraf: Fixed in da6199887cbbf7cc39414d7f1da9e16e98c7a122

vimiv/gui/metadatawidget.py
```diff
@@ -58,24 +58,24 @@ def __init__(self, parent):
...
             if self._handler is None:
-                self._handler = exif.ExifHandler(self._path)
+                self._handler = metadata.MetadataHandler(self._path)
             return self._handler
 
         @api.keybindings.register("i", "metadata", mode=api.modes.IMAGE)
         @api.commands.register(mode=api.modes.IMAGE)
         def metadata(self, count: Optional[int] = None):
-            """Toggle display of exif metadata of current image.
+            """Toggle display metadata of current image.
```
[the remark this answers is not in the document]
jcjgraf: fixed in da6199887cbbf7cc39414d7f1da9e16e98c7a122

vimiv/imutils/metadata.py
```diff
@@ -0,0 +1,440 @@
...
+                "https://karlch.github.io/vimiv-qt/documentation/exif.html",
+                once=True,
+            )
+            self._metadata = None
+
+    def fetch_key(self, base_key: str) -> Tuple[str, str, str]:
+        key = base_key.rpartition(".")[2]
+
+        if self._metadata is None:
+            return {}
```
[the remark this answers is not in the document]
jcjgraf: fixed in da6199887cbbf7cc39414d7f1da9e16e98c7a122

vimiv/imutils/metadata.py
```diff
@@ -0,0 +1,440 @@
...
+                    piexif.TYPES.Rational,
+                    piexif.TYPES.SRational,
+                ):  # (int, int) <=> numerator, denominator
+                    return (keyname, keyname, f"{val[0]}/{val[1]}")
+
+        raise KeyError(f"Key '{base_key}' not found")
+
+    def get_keys(self) -> Iterable[str]:
+        if self._metadata is None:
+            return iter([])
```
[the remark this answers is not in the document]
jcjgraf: fixed in da6199887cbbf7cc39414d7f1da9e16e98c7a122

vimiv/imutils/metadata.py
```diff
@@ -0,0 +1,440 @@
...
+            return ""
+
+        with contextlib.suppress(
+            piexif.InvalidImageDataError, FileNotFoundError, KeyError
+        ):
+            return self._metadata["0th"][piexif.ImageIFD.DateTime].decode()
+        return ""
+
+
+def check_external_dependancy(handler):
```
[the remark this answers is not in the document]
jcjgraf: fixed in da6199887cbbf7cc39414d7f1da9e16e98c7a122

vimiv/imutils/metadata.py
```diff
@@ -0,0 +1,440 @@
...
+                keyname = piexif.TAGS[ifd][tag]["name"]
+                keytype = piexif.TAGS[ifd][tag]["type"]
+                val = self._metadata[ifd][tag]
+                _logger.debug(
+                    f"name: {keyname}\
+                    type: {keytype}\
+                    value: {val}\
+                    tag: {tag}"
+                )
+                if keyname != key:
```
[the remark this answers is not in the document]
jcjgraf: Um yeah, I agree that this could be solved more efficiently. I have added it to the todo :blush:

tests/unit/imutils/test_metadata.py
```diff
@@ -0,0 +1,335 @@
+# vim: ft=python fileencoding=utf-8 sw=4 et sts=4
```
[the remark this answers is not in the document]
jcjgraf: My thought was to add more specific fixtures (ones which are not widely used) above the test cases in which they are used. But obviously I will change it to conform to the general styling rule of this project.

jcjgraf:
Thanks for catching some silly typos and mistakes and for the constructive feedback on my tests. I have written so many fixtures in order that I have great flexibility in how to combine them :upside_down_face: But you are right, I have not read the full documentation of pytest yet, else I would probably not have written a single testcases by now :smirk: 

I will do some more research on parametrisation and try to improve the test accordingly!

jcjgraf:
@karlch Am I heading into the right direction based on the changes 2fb9a4cf268ee076be9eb8ca7a8f3d31bc2d2073? 

What do you think about the fixtures `external_content`, `internal_content` and `metadata_content`? Can I leave them and use them as they are or is there a better way to deal with them? Personally, I do not see much of an advantage of removing them and re-defining all desired metadata data in each parametrisations.

karlch:
This is a nice first step, I guess this style can also be used for the `test_metadatahandler_get_keys` test case to avoid looping over the handlers.

I am perfectly fine keeping those fixtures, you could also just define a global constant for the fixed dictionaries at the top to reduce the number of fixtures passed to a testcase. I.e. for example:
```
EXTERNAL_CONTENT = {
...
}
```

Independent of the fixtures, the `try ... except AttributeError` nesting has gotten a bit out of hand in `value_match` and `value_match_piexif` :smile:

PR #329, opened 2021-01-23
Extend mark api

karlch:
The mark command has been extended with an optional action argument that allows forcing to (un-) mark the passed paths instead of toggling. When used in plugins, one would do
```python
api.mark.mark(paths, api.mark.Action.Mark)  # Force marking
api.mark.mark(paths, api.mark.Action.Unmark)  # Force unmarking
```
for the respective options. This should hopefully deal with all possible use-cases, and always emits `markdone` once done handling all paths. If only a single path should be marked, one can wrap it into a list, i.e.
```python
api.mark.mark([path])  # Toggle mark status of the single path
```

In addition we expose the simple `mark.is_marked` function as proposed.

@jcjgraf does this fulfill your expectations for the first point in #323? :blush:
As I guess the ideas come from the batchmark plugin, I took a quick look and think the `batchmark_end` part should now be doable  with something like
```python
all_marked = all(api.mark.is_marked(path) for path in selected_paths)
action = api.mark.Action.Unmark if all_marked else api.mark.Action.Mark
api.mark.mark(selected_paths, action)
```

I would wait with the exif related part to see what happens in #325, maybe even a bit later if we are not sure of the structuring yet, as once it is exposed via the api, changing the interface is no longer so easy.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -17,6 +17,10 @@ Added:
     the first icon of the current row in thumbnail mode.
 
   Thanks `@jcjgraf <https://github.com/jcjgraf>`_ for the idea!
+* The ``--action`` flag for the ``:mark`` command. ``--action toggle`` is the default
+  and keeps the previous behaviour of the mark command, i.e. inversing the mark status
+  of the passed path(s). In addition one can now use ``--action mark`` and
+  ``--action unmark`` to force (un-) marking the paths.
 
 Changed:
 ^^^^^^^^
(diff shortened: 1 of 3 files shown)


jcjgraf:
Sorry that I did not comment on #323, but thank you for looking into this. The mark api works now really flawless and indeed simplifies the code in Batchmark a lot; My code was a little clumsy :laughing: Thank you for improving the mark api!

I agree that we better wait for #325 before extending the metadata api.

PR #333, opened 2021-01-30
Undelete Failed to Restore Symlinks

jcjgraf:
The check `os.path.exists(trash_filename)` fails for broken symlinks (returns false). Therefore, vimiv was unable to undelete symlinks to images. An additional call to `os.path.islink(trash_filename)` fixes that issue.


karlch:
Thanks!

Issue #299, opened 2020-11-08
Improve handling of images from multiple directories


karlch:
In principle vimiv supports opening images from multiple directories in all sorts of ways (directly upon startup, via `:open`, via marks, via tags, ...), but the behaviour is then often inconsistent or buggy. Here are some of the issues:
- the library display the directory of the first image, not all images
- only images in the directory of the first image are tracked by the working directory handler. Therefore deleting/changing any of the other images is not communicated back to vimiv.

The expected behaviour of the library is up for discussion. The initial design was meant as a fast browser to search the filesystem and open images similar to using ranger. Open questions are:
- what should the library display if images from multiple directories are opened?
- where should going up a directory (i.e. `:scroll left` or press `h`) go in this case?

Obviously all opened images should be tracked by the working directory handler. The *how* is currently not clear though. One option proposed by @jcjgraf is to use a virtual directory which contains all opened images and track this via the working directory handler. We would need to take a look at the complexity of this, possibly it could also solve the library inconsistencies if placed at a logical position in the file hierarchy.

Branched off from #293

jcjgraf:
I have spend some time thinking about this particular issue as well as about the library mode on itself and the interplay with the other modes. Reading the docs, some issues as well as from my personal experience, I think that there is a conflicting view between what it is designed to to and what it is required to do.

If I understand you @karlch correctly from reading the docs and your perspective in some issues, library mode is designed as fast means of browsing directories and opening the desired image. Apparent from #14 all its functionality could be mimicked by the command `:open`.

Later on it was requested that additionally to opening images, library mode should also synchronize selection with other modes so that it can kind of be used to keep the overview over the images while browsing through the images in thumbnail/image view. This makes also lots of sense since unlike e.g. thumbnail mode, library mode ban be used next to image mode (have both of them open at the same time).

An then there is me coming and complaining that library mode does not properly reflect opening images from multiple directories using `:open`. :upside_down_face: 

All in all, I would say that library mode was build to open images, but in addition also required to reflect the current working set (list all open images, synchronize movements to keep an overview of the open images etc.). In my view, these two tasks cannot really be merged into a single mode in an intuitive may. 

So, if we would like to implement that in a intuitive may, we would have to separate the two functions;
[...]

PR #336, opened 2021-02-06
Add listview option in thumbnail mode

karlch:
As code speaks louder than works, here is a draft PR for what I meant with using thumbnail mode for the described listview in #299.

Thumbnail mode can now also be shown at the right of the current image as a simple list similar to the filmstrip used by other programs. In addition, the display of thumbnail mode can now be any combination of icon and text, with the default being icon-only in icon view, and icon + text in list view. The view mode is switched by the `thumbnail.listview` boolean setting. Toggling is currently bound to the `<tab>` key. Showing the icon is changed using the `thumbnail.display_pixmap` setting (should probably be renamed to `display_icon`). Showing the text is changed using the `thumbnail.display_text` setting. Each of these two setting can have one of the following options:
- never: never show the related element
- always: always show the related element (default for the icon)
- listview: only show in listview (default for text)
- iconview: only show in iconview

What do you think of the general concept @jcjgraf ?

In case we go for this, I see some more todos:
- [ ] Simplify switching between options showing text / icon, possibly using keybindings or completely refactoring the concept
- [ ] Clean up code addressing anything marked TODO and more
- [ ] Add tests
- [ ] Update documentation, probably with a screenshot for the new list view


jcjgraf:
This is golden! It adds exactly what I was missing from vimiv; a synchronized list, displaying all images opened from multiple directories, in thumbnail view next to image view.

Thank you very much for this work!

You are right, code speaks louder than words. In #299 I though I meant to integrate this view directly into the library, which would hot have solved the initial problem of #299. This implementation is somewhat similarly to what I meant. But I did not expect it to be that "easily" integrable.

I have only three things minor things to add:
- Toggling this module is bit cumbersome. While in image mode it required <kbd>gt</kbd>, <kbd>TAB</kbd>, <kbd>gi</kbd> to open it, and the same in reverse to close it again. I think a command or binding would be appropriate for that. Especially, since I am not sure if it is possible to add a custom binding for the toggling because of the required <kbd>TAB</kbd>.
- When the thumbnail listing is open and the thumbnails are rather large, it can happen that the listing overlay the image even though if the image was shifted to the left, both would fit next to each other without interference. Since this may not easily fixable, we can also leave it as it is.
- Lastly, except for the largest thumbnail size, image names are often cut of. Maybe we could split the name in two lines or something like that. Though, I don not really mind leaving it as it is as I do rather not display the filename anyways :upside_down_face:

jcjgraf:
I have noted a inconsistency in the synchronisation between thumbnail and image mode. I.e. in image mode we see image A while thumbnail listing mode shows image B as active.

### Reproduce
- Open some image in image mode (call it image A)
- Switch to thumbnail mode (normal mode, not the new list)
- Select another image (call it image B)
- Press tab to open the thumbnail listing

### Expected Behaviour
- Image mode and thumbnail listing mode both show image B

### Actual Behaviour
- Image mode shows image A
- Thumbnail listing mode shows image B

### Note
Synchronisation is respored after switching to the next image

karlch:
Thanks a lot for your detailed comments! Glad you like the general idea.

* I agree toggling is currently cumbersome. In principle one could add a custom keybinding like `enter thumbnail && set thumbnail.listview true && enter image`. Do you think it would make sense to have a default binding for this? If so, any preferences? I was thinking of `gv` as `gl` is already taken. And do you think it should always enter image mode, or should it stay in the thumbnail view?
* Understandable, but this sounds like it may be more effort than gain. If it is easy to fix, I will add it.
* Ah, yes, the issue makes perfect sense, although I propose another solution. There is the `thumbnail.font` style option and I think a much smaller default than the overall font makes sense here. We could also allow the text to be wider than the actual thumbnail. What do you think?
* The synchronization issue makes perfect sense, image mode never follows the other modes as loading images is expensive. I would not change this behavior. Instead I think it makes sense to add a flag to `open-selected` which decides whether or not thumbnail mode should be closed, similar to the library. Like this one could restore synchronization by pressing a key without leaving thumbnail mode. But I don't really have a key in mind, is there anything intuitive you can think of? `<space>` is one that came to my mind, but that is a bit of a stretch.

jcjgraf:
I am very sorry for the long delay with answering and that I have not worked on #325.

- Um, that must have been a late-evening though of myself. Obviously, one can easily create such a binding since tab itself is a binding :laughing:. I would propose `gf` for a binding since many other application call such a view "filmstrip". Personally, I thing it would be useful to end off in image mode after using this binding. Though, I can also understand when people prefer staying in thumbnail mode. Maybe it is a bit overkill, but why not adding a boolean setting which determines if the binding is `enter thumbnail && set thumbnail.listview true` or `enter thumbnail && set thumbnail.listview true && enter image` respectively?
- No worries, I am fine with how it is...
- That is a possible solution too, though I am not sure if it is sufficient to decrease `thumbnail.font` size to make filenames of "average length" fit. Allowing the text to overflow the thumbnail box may work in filmstrip view, but when the image names are displayed in the thumbnail view, image names may easily overlap. We could certainly allow the text to be wider than the thumbnail in filmstrip view only, though I am not sure it this would look neat anyways. I am fine with slightly decreasing the `thumbnail.font` size and else, leave as it is. We can still solve this somehow, sometimes in the future if it turns out to be an issue in the long run.
- My expected synchronisation was that when I select an image in thumbnail mode and then open the thumbnail listing (press TAB), that the image displayed in image mode is the one I selected in thubmanil mode. I have attached a GIF which hopefully helps to show you my point.

VimivSynchronisation

karlch:
Absolutely no worries, I am also currently rather busy and will probably not find the time to work on this in the next few weeks unfortunately.

- `gf` sounds good to me, but I am not a fan of having a setting for the entering or not. How about having two bindings, e.g. `gf` and `gF` where `gf` would just enter the filmstrip mode (which is more consistent with the other `gx` bindings) and `gF` would open it and then enter image mode. Maybe there is also a better binding combination than this.
- Concerning the fontsize, let's see how this goes, your concerns are all valid and this will require some playing around with the various use-cases.
- I am not quite sure how to best synchronize here. I guess loading the currently selected image when entering "filmstrip" mode could be a valid option. But the image would certainly not "follow" any scrolling in filmstrip mode by default, much like it doesn't follow the library selection unless explicitly requested. I guess we should add `n` and `p` here as well as a binding to open the selected image without leaving filmstrip mode.

jcjgraf:
- I like `gf` and `gF`. I have also thought that we could create just one binding depending if the current mode is thumbnail or image, the binding would end in thumbnail or image mode respectively. Though, separate bindings may be more powerful and less confusing.
- :white_check_mark: 
- This makes all sense. I think loading the currently selecting image when opening the film strip would mostly solve the issue I guess. I am not sure what you mean with adding the bindings `n` and `p`.

karlch:
All right, this took a while but finally there is some more progress :blush: 

- The two bindings have been added, although `gF` ends with `open-selected` instead of `enter image` to partially address the synchronization issue.
- There was quite some progress on the rendering, mainly the height and width of thumbnail elements now adapt to the visible content. Let me know if you encounter anything weird, this was quite some havoc :laughing: 
- I have added `n` and `p` in thumbnail mode, you can best test them when in filmstrip view (and in thumbnail mode). I am still not convinced we should display the corresponding image immediately, but may just go for it, although not tonight :smile: 

In addition to all this, the settings have changed in the behaviour, i.e. we now set "what is visible in mode x" and not "when is x visible". This has a few advantages as described in 86219bd1b427ae254da02a97c83b46546f8ef8cc. Mainly, it is now possible to jump through the different display options. This is currently bound to `w` without any particular reason except for it being easy to type and close to `<tab>`. Let me know if you can think of a meaningful binding :blush:

jcjgraf (approved):
Thanks for you late-evening shift, I like a lot what you have done :+1: 

I have gone through you changed. While I did not understand every single line (esp. QT related things), I have still added a few comments. Two of them indicate bugs which happen only when you use rather long file names. I hope I have described everything well enough, I you do not understand something, let me know.

Additionally, I think we have a little mess with the naming of all these modes/view. I.e. we have thumbnail mode, icon mode, list mode, filmstrip view, (grid mode) as well as combinations with other modes (image mode + filmstrip view).
I presume that this can be rather confusing for users (at least it is for me). While from the code base, icon mode and filmstrip mode are similar (both depend on the thumbnail mode) they are in practice different things. I do rather see the filmstrip mode as an additional view of image mode instead of a view of thumbnail mode.

I think a layer of abstraction would here be appropriate; Image mode has like three views:
- Normal (like image mode before this PR)
- Filmstrip open (filmstrip open but image mode activated)
- Filmstrip active (filmstrip open and active)

Image mode is entered using *gi*. From there, one can toggle between *Normal* and *Filmstrip open* view using *tab*. From any mode (/view) *Filmstrip active* can be accessed using *gf*. When using *gi* from any mode, we get to *Image Normal* or *Image Filmstrip open* depending on the toggle of filmstrip (i.e. depending on `thumbnail.filmstrip`).

Thumbnail mode would be equivalent to how it was before this PR (with the exception of the option to display the image name).

As said, while this change may not make sense from the POV of the code base, I belief it improves the user experience. Let me know what you think.

Issue #338, opened 2021-02-17
Support for .avif format


schyzophrene-asynchrone:
Hi!

I recently discovered a new image format: AVIF.

There is a libavif in most distribution repositories, however installing it on Arch Linux did not allow me to open the following file:
https://linkmauve.fr/files/railgun.avif

Support for this would be nice. ;)

Thanks again for your software.

schyzophrene-asynchrone:
As for the previous issue I opened, if you indicate to me what I should do, I can take care of it. It may take a while, but I guess no one is in a hurry regarding this.

karlch:
Hi!

Thanks for reporting this! I found [this qt imageformats plugin](https://aur.archlinux.org/packages/qt-avif-image-plugin-libavif-git/) which should make it possible for qt to read and write avif files. Unfortunately avif is not recognized by [imghdr](https://docs.python.org/3/library/imghdr.html) which we use to identify the image type of any file.

The easiest way forward should be to:
- install the AUR plugin
- add a test that identifies if a file is of type `avif`

We already have `vimiv/plugins/imageformats.py` to handle this usecase so it should be fairly straightforward to add a test similar to the one we have for `cr2`, as long as one can identify `avif` files using the bytes header.

Hope this helps! Let me know if there is anything I can help with :blush:

PR #343, opened 2021-02-20
Add support for detecting .avif files using imageformats plugin

schyzophrene-asynchrone:
I have not tested this, nor did I find any documentation regarding the bytes header of an avif file.

What I did was downloading images from https://github.com/AOMediaCodec/av1-avif and testing them with `hexdump -n 16 -C`. All of them, except for two with animated features, have a `ftypavif` in byte range 4-12. The two animated have a `ftypavis` in the same range.

[It would seem](https://stackoverflow.com/a/66239407) that some avif files could have another bytes header, but I could not find any relevant document about this.

## Related Issues
#338

### DIFF
--- a/vimiv/plugins/imageformats.py
+++ b/vimiv/plugins/imageformats.py
@@ -38,7 +38,7 @@ def test_cr2(header: bytes, _f: Optional[BinaryIO]) -> bool:
 
 
 def test_avif(header: bytes, _f: Optional[BinaryIO]) -> bool:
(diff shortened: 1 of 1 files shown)


karlch:
Awesome, thank you, I will merge this once I fix the new CI havoc :smile:

karlch:
Finally found the time to play around with the avif file format. Seems like everything works fine, even the animations, when changing the mentioned typo. I will update this and the changelog after merging. Thanks again!

Issue #358, opened 2021-04-06
Is it possible to save the contents of an image to the clipboard?


pglira:
Often it would be useful to copy the contents of an image in order to paste it, e.g. in Teams, WhatsApp, LibreOffice, ... I haven't found an option/command for this. Can it be added via config?

jcjgraf:
There is not such functionality in vimiv yet. But in #359 I have added the command `copy-image` which should do what you are asking for. It would be great if you could have a look at it and check if it works and does what you want.

pglira:
Thanks!

I've installed the new version and can see the new options.

However, it seems that it is not always working. So far I haven't found out if it is related to the image format (png, jpg, bmp) or to the target program, e.g. I can paste the image often in libre office writer, but almost never in whatsapp web.

jcjgraf:
A more extensive tests confirms your findings that not all images and all target programs works equally well.

I think it is mainly related to the target program and file size. It seems like some programs are not able to handle images larger than some size. For me, copying to Libre Office Write works for all jpgs of any size (tested up to 8 MB), while Element (Riot) only seems to allow pasting images smaller than 1MB. I could imagine that the threshold for web applications may be even lower.
Currently, I do not have any other image formats available, but I think this should generally works for all file types which vimiv supports out of the box (it certainly does not work for CR2 yet).

Could you try to confirm these findings by e.g. trying to copy rather small images to whatsapp web? Do you have an other application which allows copy/pasting images? If so, are there images which you can copy/paste to Whatsapp Web from the other application but not for vimiv?

karlch:
Just merged #359 which implements this along with an option to reduce the image size before copying. All credit goes to @jcjgraf, thanks a lot for working on this!

@pglira let us know if you encounter further issues / have other ideas for improvement :blush:

PR #359, opened 2021-04-06
Copy Image to Clipboard

jcjgraf:
Adds a new command `copy-image` to copy the whole image pixmap to the system clipboard, allowing easy pasting to other applications. `yi` and `yI` are currently bound to this command (global clipboard and selection clipboard respectively). 

Todo:
- [x] Check if implementation fulfils requirements
- [ ] ~Flag to copy the original image in contrast to the (potentially compressed) QPixmap~
- [x] Unit tests?

Implements: #358

### DIFF
--- a/tests/end2end/features/misc/clipboard.feature
+++ b/tests/end2end/features/misc/clipboard.feature
@@ -28,26 +28,26 @@ Feature: Interaction with the system clipboard.
         Then the working directory should be child_01
 
     Scenario: Copy image to clipboard.
-        Given I open the image '.hidden.jpg'
+        Given I open any image
         When I run copy-image
-        Then the clipboard should contain image .hidden.jpg
+        Then the clipboard should contain any image
 
     Scenario: Copy image to primary.
-        Given I open the image '.hidden.jpg'
+        Given I open any image
         When I run copy-image --primary
-        Then the primary selection should contain image .hidden.jpg
+        Then the primary selection should contain any image
 
     Scenario: Copy image to clipboard and scale width.
-        Given I open the image '.hidden.jpg'
+        Given I open any image
         When I run copy-image --width=100
         Then the clipboard should contain an image with width 100
 
     Scenario: Copy image to clipboard and scale height.
-        Given I open the image '.hidden.jpg'
+        Given I open any image
         When I run copy-image --height=100
         Then the clipboard should contain an image with height 100
 
     Scenario: Copy image to clipboard and scale size.
-        Given I open the image '.hidden.jpg'
+        Given I open any image
         When I run copy-image --size=100
         Then the clipboard should contain an image with size 100
--- a/tests/end2end/features/misc/clipboard.feature
+++ b/tests/end2end/features/misc/clipboard.feature
@@ -26,3 +26,28 @@ Feature: Interaction with the system clipboard.
         When I run copy-name
         And I run paste-name
         Then the working directory should be child_01
+
+    Scenario: Copy image to clipboard.
+        Given I open the image '.hidden.jpg'
+        When I run copy-image
+        Then the clipboard should contain image .hidden.jpg
+
+    Scenario: Copy image to primary.
+        Given I open the image '.hidden.jpg'
+        When I run copy-image --primary
+        Then the primary selection should contain image .hidden.jpg
+
+    Scenario: Copy image to clipboard and scale width.
+        Given I open the image '.hidden.jpg'
+        When I run copy-image --width=100
+        Then the clipboard should contain an image with width 100
+
+    Scenario: Copy image to clipboard and scale height.
+        Given I open the image '.hidden.jpg'
+        When I run copy-image --height=100
+        Then the clipboard should contain an image with height 100
+
+    Scenario: Copy image to clipboard and scale size.
+        Given I open the image '.hidden.jpg'
+        When I run copy-image --size=100
+        Then the clipboard should contain an image with size 100
--- a/tests/end2end/features/misc/conftest.py
+++ b/tests/end2end/features/misc/conftest.py
@@ -4,7 +4,7 @@
 # Copyright 2017-2021 Christian Karl (karlch) <karlch at protonmail dot com>
 # License: GNU GPL v3, see the "LICENSE" and "AUTHORS" files for details.
 
-from PyQt5.QtGui import QGuiApplication, QClipboard
+from PyQt5.QtGui import QGuiApplication, QClipboard, QImage
 
 import pytest
 import pytest_bdd as bdd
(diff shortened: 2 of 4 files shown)


wedens:
I've tried copying using the new command and pasting to several apps (some of them are: discord, pinta, rambox, qutebrowser) and it works as expected.

karlch:
Thanks a lot for the PR and sorry for being inactive recently, let's just say life is currently a bit weird :laughing: 

The implementation LGTM, what would remain is a quick test and an update to the changelog, but I am happy to look into this after merging if you prefer. I should find quite some time to work on vimiv over the week-end :blush:

jcjgraf:
Thanks @wedens for your testing and feedback!

As pointed out by @pglira and briefly discussed in #358, copying does not work for some images and target applications. I have the feeling that this is related to limitations in image size of the target application and hence nothing which we can fix.

Currently, the QPixmap opened by vimiv, is copied and **not** the original images file. Therefore, metadata may be lost and the image quality may also deteriorate.

Therefore, I thought it would be a good addition to allow copying the original file through a flag for `copy-image`. I have come up with the following snippet:

```Python
url = QUrl.fromLocalFile(path)
mime_data = QMimeData()
mime_data.setUrls([url])
clipboard.setMimeData(mime_data, mode=mode)
```

In (scientifically meaningless) tests (due to the lack of applications which allow image pasting) this method only worked for Libre Office Write and not for Element (Riot) :thinking:  Therefore, I am unsure how widely supported this method is.

Unless someone has a solution (or can confirm that this method actually works well in other applications) I would suggest to stick to the current implementation and potentially open an issue to remind us to add this functionality some point in the future.

What do you reckon @karlch @pglira?

karlch:
I think I prefer the current implementation. If one really wants all the information of the current image in its original state, one can always copy the abspath and deal with it in the other application accordingly. I would assume this command would be used more for simple inserting of the image and it would probably be resized (libreoffice) or compressed (web applications) afterwards in any case. If you assume the actual image data size is an issue, could it make sense to add flags for `--width` and `--height` to rescale the `QImage` before copying? This could also come in handy to not have a huge image inserted into libreoffice. What do you think?

jcjgraf:
Thanks for the feedback and this great idea!

I have added the flag `width` and `height` to scale the image to the specified with **OR** height respectively. It is not possible to scale it to a specified wight and hight. Should that be possible?
I have also added the `size` flag which scales the longer side of the image to the specified value.

There is a hardcoded precedence if the user specifies multiple flags. Should an error or something be thrown in such a case or is it alright how it is implemented?

Let me know if you have a neater way to implement this! Also tell if you prefer the code to be more compact (i.e. less empty lines :upside_down_face: )

jcjgraf:
I have tried to put some meaningful end2end tests for this new command together. For some reason the CI is not happy with them even though they work for me locally :confused: 

I am not very familiar with writing test, so feedback is very welcome :blush:

One question I have is what images can I open with `I open the image <image>`? I have just selected `.hidden.jpg` since this is the only one you have used in this context.

tests/end2end/features/misc/clipboard.feature
```diff
@@ -26,3 +26,28 @@ Feature: Interaction with the system clipboard.
         When I run copy-name
         And I run paste-name
         Then the working directory should be child_01
+
+    Scenario: Copy image to clipboard.
+        Given I open the image '.hidden.jpg'
```
karlch, line 31: This `Given` step allows defining the name of the image we open, so you can use any name you like. If you don't need a specific name, you can just use `Given I open any image` which will create `image.jpg`.

tests/end2end/features/misc/conftest.py
```diff
@@ -23,3 +23,33 @@ def check_clipboard(clipboard, text):
 @bdd.then(bdd.parsers.parse("The primary selection should contain {text}"))
 def check_primary(clipboard, text):
     assert text in clipboard.text(mode=QClipboard.Selection)
+
+
+@bdd.then(bdd.parsers.parse("The clipboard should contain image {image}"))
+def check_clipboard_image(clipboard, image):
+    assert clipboard.pixmap(mode=QClipboard.Clipboard).toImage() == QImage(image)
```
karlch, line 30: I asssume this test fails for some reason under qt 5.9 and 5.10. I would be happy to loosen this to `should contain any image` where we just ensure that we successfully saved an image to the clipboard without comparing the content directly.

tests/end2end/features/misc/conftest.py
```diff
@@ -23,3 +23,33 @@ def check_clipboard(clipboard, text):
...
+
+
+@bdd.then(bdd.parsers.parse("The primary selection should contain image {image}"))
+def check_primary_image(clipboard, image):
+    assert clipboard.pixmap(mode=QClipboard.Selection).toImage() == QImage(image)
```
karlch, line 35: Same as above :blush:

vimiv/api/_modules.py
```diff
@@ -89,6 +89,54 @@ def copy_name(abspath: bool = False, primary: bool = False) -> None:
     clipboard.setText(name, mode=mode)
 
 
+@api.keybindings.register("yi", "copy-image")
+@api.keybindings.register("yI", "copy-image --primary")
+@api.commands.register()
+def copy_image(
+    primary: bool = False, width: int = -1, height: int = -1, size: int = -1
```
karlch, line 96: I would personally prefer to use `None` instead of `-1` as it fits better to the `Optional` character of this flag.

Another random idea I had is to support `count` here which would set the size using `count` so e.g. `800yi` would result in the same behaviour as `:copy-image --size 800` to allow setting the size without interacting with the command line.

karlch (commented):
Thanks a lot, this looks great! I have added a few minor comments up for discussion and to clarify the CI :blush:

tests/end2end/features/misc/conftest.py
jcjgraf: This seems to work. Changed that in 0fb4369

tests/end2end/features/misc/conftest.py
jcjgraf: Changed in 0fb4369

vimiv/api/_modules.py
jcjgraf: Imrpoved the None in cdbacf4

Added the count support in d767e5b55d45f14ba22f5cde86ad00f337d3c4a4. Have I added the documentation for the count correctly to the docstring?

karlch: Looks good to me except for possibly a missing whitespace after `--size`.

vimiv/api/_modules.py
```diff
@@ -89,6 +89,62 @@ def copy_name(abspath: bool = False, primary: bool = False) -> None:
...
+        return
+
+    if size or count:
+        pix_size = pixmap.size()
+
+        size = count if count else size
+
+        if pix_size.height() >= pix_size.width():
+            _logger.debug(f"Copy image with size {size} restricting height")
+            pixmap = pixmap.scaledToHeight(size)
```
karlch, line 132: Feel free to add `  # type: ignore[arg-type]` to the two lines failing mypy. It probably cannot deal with this slightly non-trivial check to know that size is now certainly an `int` and no longer possibly `None`.

With this and splitting the syntax section into two lines
```
    **syntax:** ``:copy-image [--primary] [--width=WIDTH] [--height=HEIGHT]
    [--size=SIZE]``
```
the nasty CI should finally be happy (except for the annoying timeout issue) :laughing:

vimiv/api/_modules.py
jcjgraf: good catch, fixed in 421e0c65c372d0672455c22bcfd9a32d602e16e4

vimiv/api/_modules.py
jcjgraf: All green, thanks for the hints!

PR #360, opened 2021-04-06
Wrong URL to pyexiv2 project

jcjgraf:
I have just noted that the link to the pyexiv2 project in the documentation is actually wrong. The reason being is that there are (at least) three projects called pyexiv2 (or very similar). We are using [this](https://python3-exiv2.readthedocs.io). I have fixed the URL and added a note to the docs that one has to take care to install the right package.

~I am not sure how to compile and preview the docs locally. So, I cannot guarantee that I did everything right with the warning.~ **Edit:** I just should looked at you contribution guideline in the first place :see_no_evil: 

@karlch feel free to rewrite and change things you do not like.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -56,7 +56,7 @@ Added:
 * New ``--keep-zoom`` flag for ``:next`` and ``:prev`` which preserves zoom level and
   scroll position of the current image.
   Thanks `@jcjgraf`_ for the idea!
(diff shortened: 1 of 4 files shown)


karlch:
Thanks for catching this, the packages are indeed quite confusing. This looks good on a first glance, I will build the documentation and merge this along with the other open PRs over the week-end :blush:

Issue #368, opened 2021-04-21
[Feature] Add view (read-only) mode


willemw12:
It would be nice to have a view mode (read-only mode) and not have to worry about accidental overwriting of files during that vimiv session.

In view mode, all commands that change or (re)move an image (buffer) are disabled.

To run in view mode, run: `vimiv -R`. It is not necessary, or maybe not even desired, to have it as a config setting.

As an alternative to view mode, one can remove the mappings in the keyfile that change or (re)move images.

karlch:
Thanks for opening an issue on github!

This sounds do-able, I will look into it :blush:

karlch:
Implemented in #369.

willemw12:
Vim appends " [RO]" to the filename on the status line when in view mode. Maybe vimiv can do the same. Or append " [RO]" to the mode on the status line.

karlch:
This makes a lot of sense. I went ahead and added a corresponding status module in 4124df891ca9bc14a0dfab4724567e8366c46565. Let me know if this works as expected :blush: 

Note that you probably have to update your `vimiv.conf` as described in the [changelog](https://karlch.github.io/vimiv-qt/changelog.html).

willemw12:
It all works correctly.

PR #369, opened 2021-04-21
Add read only setting

karlch:
As described in #368 this adds a new `read_only` setting. When active, all commands that may write changes to disk are disabled.

I decided to implement this as a setting instead of a plain parser flag, as settings are nicely available everywhere using the existing settings api. I have to admit that `vimiv -s read_only true` or `vimiv -s read_only 1` is certainly more typing than a simple `vimiv -R` though. @willemw12, do you think the flag should be added in addition? I usually prefer not to do this, as it becomes difficult to decide which settings should also have parser flags, and which should not.

Another open question is how to deal with shell commands. Currently both `:!` and `:spawn` are available when `read_only` is active. This is somewhat borderline as they can obviously make changes to disk. On the other hand, I don't ever see this happening accidentally, and there are clearly also commands that do not write to disk. Any thoughts?

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -39,6 +39,9 @@ Added:
 * The ``:copy-image`` command which copies the selected image to the clipboard. The
   flags ``--width``, ``--height`` and ``--size`` allow to scale the copied image.
   Thanks `@jcjgraf`_!
+* The ``read_only`` setting. When set to true, all commands that may write changes to
+  disk are disabled.
+  Thanks `@willemw12`_ for the idea!
 
 Changed:
 ^^^^^^^^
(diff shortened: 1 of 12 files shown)


willemw12:
Thanks. That was very fast.

Reusing settings instead of adding a dedicated CLI flag is fine. It is still less typing if defined as a shell alias or script, thanks to tab-completion:

    alias vimiv-view='vimiv -s read_only true'

Regarding shell commands: the same applies to vi/vim. You can still run shell commands from vi/vim when in view (read-only) mode.

karlch:
Nice, I will go ahead and merge this then :blush:

PR #373, opened 2021-04-26
Mention new Video plugin in docs

jcjgraf:
I have written a little plugin which adds *Video Support* to vimiv. Videos are listed within vimiv (this way i do not miss then when copying images from my SD card using vimiv :blush:) and can be played using an external player like MPV or something.

At the time, only QuickTime Movie (`.mov`) is supported but I intend to add new formats soon.

P.S. is there a particular reason why in the docs the default plugins are listed in a table and the user plugins in a list? I think a table looks much nicer.


karlch:
Thanks a lot! I haven't taken a detailed look at the plugin yet, but the first frame is displayed as expected, cool stuff!

I can't think of any reason except for me messing up at the time. I will change the user plugins to a table after merging, thanks for noticing :blush:

Issue #136, opened 2019-12-03
Default mouse click bindings


karlch:
I personally don't use the mouse much, so I also don't know what sane mouse bindings for various widgets would be.

Library and thumbnail mode seem pretty much set. Maybe binding `<button-right>` in the library to `:scroll left` a.k.a. go to the parent directory, would make sense.

In image mode we may want to:
* Focus the next/previous
* Switch to thumbnail mode
* Switch to library mode
* ...

Input from anyone using the mouse more than me would be very helpful :smiley:

BachoSeven:
Regarding the library bindings, `<button-right>` for parent directories makes sense to me; I also use that in `lf` for instance.

For image mode, I would suggest having `<button-right>` bound to switching to Library mode, since this is muscle memory I have from using `sxiv`.

Although most people only have a few mouse buttons, the most useful bindings for focusing the next/previous image are what `xev` reports as `Button 8` and `Button 9`, present on many mice as "Previous/Next" Buttons (i.e. used for navigating tab history in a browser). This could perhaps be configurable(?) 

Alternatively, one could use the scrolling wheel to focus the next image, but I think the scrolling wheel would be most suited to zooming in/out in Image mode.

karlch:
Thanks a lot for your input, all of what you say makes sense to me. I will see if I can find a mouse with the corresponding buttons and test the next / prev stuff with them. Once done, I guess adding:
- `<button-right>` for parent directory in library mode
- `<button-right>` for entering the library in image mode
- `<button-8>` / `<button-9>` for switching images in image mode

as default bindings can just be done.

Note that mouse bindings are configurable in the `keys.conf` like any regular keys. A nice way to figure out the names of the bindings in vimiv is by launching vimiv with `--debug gui.eventhandler` as it spits out the event like:
```
[14:55:43] DEBUG    <gui.eventhandler>   EventHandlerMixin: handling <button-right> for mode image
```

I would also personally not touch the scrolling wheel for switching images. Like you say, it makes sense to use it for zooming. This is currently doable when using it while pressing `<ctrl>`. Without the modifier, it is used for scrolling the image in case it is zoomed.

BachoSeven:
Thanks, I didn't notice `<ctrl>`+scroll-wheel was already implemented.

karlch:
[...]

PR #378, opened 2021-05-01
Change mouse scroll behaviour in library

karlch:
As this has come up both in #136 and in the recent youtube video, I went ahead and changed the behaviour of scrolling with the mouse in the library. Instead of just scrolling the view, it now changes the selection. Personally I have no preference, as I almost never use the mouse, and felt one can always click wherever one wants when scrolling the view.

@jcjgraf @BachoSeven any thoughts if this version would be preferable?

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -59,6 +59,8 @@ Changed:
 * New set of default metadata key sets numbered from 1 to 5. Thanks `@jcjgraf`_!
 * When toggling ``library.show_hidden`` the selection now stays on the same file, not
   the same index.
+* Using the mouse to scroll in the library now changes the selection instead of just
+  scrolling the view.
 
 Fixed:
 ^^^^^^
--- a/vimiv/gui/library.py
+++ b/vimiv/gui/library.py
@@ -347,6 +347,14 @@ class Library(
         if open_selected_image and not os.path.isdir(current):
             self.open_selected(close=False)
 
+    def wheelEvent(self, event):
+        """Update mouse wheel for proper scrolling."""
+        steps = int(event.angleDelta().y() / 120)
+        if steps < 0:
+            self.scroll(argtypes.DirectionWithPage.Down, count=abs(steps))
+        else:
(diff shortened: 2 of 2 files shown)


BachoSeven:
Yes, I think it's a bit more intuitive. This way, we would have the Backward/Forward button behaving as `n`/`p`, and scrolling up/down analogous to `j`/`k`, which makes sense to me.

BachoSeven:
What I mean is that the current behaviour of mouse scroll had no "correspondent"  keyboard binding, making it perhaps more unexpected.

jcjgraf:
I do not have much "mouse-experience" either, but I think this way it is definitely more consistent with how other applications use the scrolling.

How did you come up with the hardcoded scroll factor? For me it is way to large. About 5 - 10 works best for me (instead of 120).

karlch:
@jcjgraf I assume you use a touchpad instead of a physical mouse? I had the same issue with 120 being very large for my rather sensitive touchpad, but it is the default for a regular mouse wheel as [described in the Qt documentation](https://doc.qt.io/qt-5/qwheelevent.html#angleDelta) and indeed when I use my regular mouse one "roll" ends up doing a single step in the library like this.

jcjgraf:
I am indeed using the trackpad. It is somewhat funny that this value must be so different for different mouses especially since scrolling in thumbnail view works just fine as it is.

karlch:
I assume the difference was that we only scrolled the library in "integer steps" whereas in the other areas we can support arbitrary floats. To hopefully fix this I went ahead and used the option to "cumulatively add the delta values from events" as described in the docs in 45a1ef4aacf28c1d26d672c735f01f25fd7d8b9b. This is getting quite involved for mouse support, but now the library also behaves nicely on my trackpad. Let me know if you still experience trouble :blush:

jcjgraf:
This is great, works as expected. Just make sure you get enough sleep too! :zzz:

karlch:
Sounds good, I will go ahead and merge this then.

Interesting way to spell coffee, haven't seen that one before :smile:

Issue #379, opened 2021-05-01
Printing marked images to stdout


loiccoyle:
Is there a way to print the marked images to stdout similar to `sxiv`? I would like to use `vimiv` in some scripts.

jcjgraf:
Thanks for your interest in vimiv!

I can see the power of such a command, but as far as I know there is not such functionality yet. I quickly went forward and added the command `:mark-print` in #380. Would you mind to check it out and tell us if it fulfils your requirements?

I do not think that you are looking for that, but you can also launch shell scripts from within vimiv and pass a list of all marked images as augments to the script. The syntax is the following: `:!myScript %m`. For full documentation see [here](https://karlch.github.io/vimiv-qt/documentation/getting_started.html#command-line).

loiccoyle:
Ideally I would like to be able to do something like: 
```sh
SELECTED_PICTURE="$(find pictures/ -type f | vimiv)"
```
Where `vimiv` opens up in thumbnail mode, I can then select a single or multiple images, when I quit the marked images are printed to stdout.

`sxiv` handles this usecase with launch arguments, [see man page](https://www.mankier.com/1/sxiv): 
```sh
SELECTED_PICTURE="$(find pictures/ -type f | sxiv -tio)"
```

Maybe something similar could be implemented in `vimiv`?

karlch:
In principle this should be doable after we merge #380 with a workflow along:
```bash
SELECTED="$(vimiv $(find pictures -type f) --command 'enter thumbnail'))"
```
and running the command defined in the PR before closing vimiv. This could of course be simplified / shortened by defining aliases or keybindings accordingly. Is this going in the right direction?

Personally I have found running scripts from vimiv and passing the marked images as described by @jcjgraf the easier workflow. Everything more complicated I like solving using a dedicated [plugin in python](https://karlch.github.io/vimiv-qt/documentation/configuration/plugins.html).

jcjgraf:
> Personally I have found running scripts from vimiv and passing the marked images as described by @jcjgraf the easier workflow.

While I agree with that, the workflow @loiccoyle  is describing is in essence *rofi for images*. And I think that is especially powerful if is is nicely and easily usable from within shell scripts.

loiccoyle:
> While I agree with that, the workflow @loiccoyle is describing is in essence _rofi for images_. And I think that is especially powerful if is is nicely and easily usable from within shell scripts.

Yes exactly !

[...]

PR #380, opened 2021-05-02
Command to print to STDOUT

jcjgraf:
Adds the new command `print-stdout` to print to STDOUT from within vimiv, and an alias to print all marked images.

~Implements #379~

### DIFF
--- a/tests/end2end/features/api/print.feature
+++ b/tests/end2end/features/api/print.feature
@@ -1,0 +1,42 @@
+Feature: Print to STDOUT
+
+    Scenario: Print trivial list
+        Given I start vimiv
+        And I capture output
+        When I run print-stdout "test"
+        Then stdout should contain 'test'
+
+    Scenario: Print longer list
+        Given I start vimiv
+        And I capture output
+        When I run print-stdout "test1" "test2"
+        Then stdout should contain 'test1'
+        And stdout should contain 'test2'
+
+    Scenario: Print list using sep
+        Given I start vimiv
+        And I capture output
+        When I run print-stdout "test1" "test2" --sep ','
+        Then stdout should contain 'test1,test2'
+
+    Scenario: Print list using end
+        Given I start vimiv
+        And I capture output
+        When I run print-stdout "test1" "test2" --end 'xxx\n'
+        Then stdout should contain 'test2xxx'
+
+    Scenario: Print marked images
+        Given I open 2 images
+        And I capture output
+        When I run mark image_01.jpg image_02.jpg
+        And I run print-stdout %m
+        Then stdout should contain 'image_01.jpg'
+        And stdout should contain 'image_02.jpg'
+
+    Scenario: Print marked images using alias
+        Given I open 2 images
+        And I capture output
+        When I run mark image_01.jpg image_02.jpg
+        And I run mark-print
+        Then stdout should contain 'image_01.jpg'
+        And stdout should contain 'image_02.jpg'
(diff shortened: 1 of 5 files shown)


karlch:
While I see the use of a command that prints the marked images, I would prefer to keep this more general, maybe something like:
```python
@api.commands.register()
def needs_name(text: List[str], sep: str = "\n", end: str = "\n"):
    """Print text to the terminal."""
    print(*text, sep=sep, end=end)
```
which could print the marked  images just by calling `:needs-name %m`. If we like, we could even add an alias for printing the marked images.

jcjgraf:
I this is a great generalisation idea. I have improved the command as you have suggested and also added an alias for printing the marked images. I have also added a series of end2end test cases.

I have named the command `:print-stdout`. If this is too long or something, I would go for `:print-out`.

The alias is named `:print-mark`, tho I am unsure which of `:print-mark` and `:mark-print` is better suited. The advantage of the later is that it is more nicely integrated into the mark commands.

Let me know if you spot anything to add or improve.

tests/end2end/features/api/print.feature
```diff
@@ -0,0 +1,42 @@
...
+        Then stdout should contain 'test'
+
+    Scenario: Print longer list
+        Given I start vimiv
+        And I capture output
+        When I run print-stdout "test1", "test2"
+        Then stdout should contain 'test1'
+        And stdout should contain 'test2'
+
+    Scenario: Print list using seq
```
karlch, line 16: seq -> sep :blush:

tests/end2end/features/api/print.feature
```diff
@@ -0,0 +1,42 @@
...
+    Scenario: Print trivial list
+        Given I start vimiv
+        And I capture output
+        When I run print-stdout "test"
+        Then stdout should contain 'test'
+
+    Scenario: Print longer list
+        Given I start vimiv
+        And I capture output
+        When I run print-stdout "test1", "test2"
```
karlch, line 12: I am not sure why there is a `,` between the two, but it doesn't do any harm of course :smile:

vimiv/api/__init__.py
```diff
@@ -93,3 +93,19 @@ def add_external_format(
     """
     files.add_image_format(file_format, test_func)
     imagereader.external_handler[file_format] = load_func
+
```
karlch, line 96: Sorry for being a bit of a nuisance here, but personally I would not expose a command that simply wraps the builtin print function via the api. Instead this could just go into `api._modules` as is :blush:

vimiv/gui/metadatawidget.py
```diff
@@ -123,7 +123,7 @@ def metadata_list_keys(self, n_cols: int = 3, to_term: bool = False):
 
             keys = sorted(set(self.handler.get_keys()))
             if to_term:
-                print(*keys, sep="\n")
+                api.print_stdout(keys)
```
karlch, line 126: As I would not expose the command as mentioned before, this would just have to be redone...

karlch (commented):
Thanks a lot, the tests look great! The only issue I have is the one concerning the exposure of the command via the api. Let me know if you have a different opinion here :blush:

tests/end2end/features/api/print.feature
jcjgraf: fixed in b039b82bcb453c33673e671b3d9d2ef7c306f7fa

tests/end2end/features/api/print.feature
jcjgraf: you're right. That should not be there. Fixed in b039b82bcb453c33673e671b3d9d2ef7c306f7fa

vimiv/api/__init__.py
jcjgraf: I thought that if we already have a *generalized* printer, everyone should also use it. But I also do not mind changing it.

Fixed in b039b82bcb453c33673e671b3d9d2ef7c306f7fa

karlch: Thanks a lot, this looks great now :blush:

PR #381, opened 2021-05-03
Vimiv as pipe

karlch:
After better understanding the idea behind #379 I went ahead and added the `-i` and `-o` flags. The described workflow with "vimiv as pipe" to select images should now be possible using:
```bash
SELECTED=$(find images -type f | vimiv -i -o "%m" --command "enter thumbnail")
```

Any feedback would be welcome @jcjgraf and @loiccoyle as I don't really use this type of a workflow myself :blush:

TODO:
- [x] Tests
- [x] Think about paths with special characters, e.g. whitespace. Currently they are quoted as usual, e.g. `'path with space'`

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -55,6 +55,9 @@ Added:
 * The ``:print-stdout`` command which prints a given list of string to the STDOUT.
 * The ``:mark-print`` default alias which prints all marked images to STDOUT using
   ``:print-stdout``.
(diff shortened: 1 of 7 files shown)


loiccoyle:
Perfect, it works nicely !

jcjgraf:
This works well for me too. I have also quickly tested special characters like whitespaces, quotes etc. and I did not notice any flaws.

karlch:
Awesome, thanks for the quick feedback, merging :blush:

PR #383, opened 2021-05-03
Adopt mouse scroll behaviour of library mode for thumbnail mode

karlch:
I feel like this is a bit of a no-brainer, but missed this somehow when implementing the new scroll behaviour in the library in #378. Thumbnail mode now reflects these changes as well and also changes the selection when scrolling with the mouse.

Any thoughts or objections @jcjgraf @BachoSeven?

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -65,8 +65,8 @@ Changed:
 * New set of default metadata key sets numbered from 1 to 5. Thanks `@jcjgraf`_!
 * When toggling ``library.show_hidden`` the selection now stays on the same file, not
   the same index.
-* Using the mouse to scroll in the library now changes the selection instead of just
-  scrolling the view.
+* Using the mouse to scroll in library and thumbnail mode now changes the selection
+  instead of just scrolling the view.
 
 Fixed:
 ^^^^^^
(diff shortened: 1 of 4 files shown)


jcjgraf:
This works really well on my trackpad. Though, without the possibility for horizontal scrolling, this new behaviour is somewhat useless, since in many cases I still have to click on the image to select it.

Also, I am actually not sure if I prefer the new or old behaviour. Thinking about this, if not too hard to implement, a setting option to select between the two behaviours may be a good addition?

karlch:
Sorry, I somehow lost track of this :disappointed: 

I agree that it is not a full replacement as one still has to click for horizontal movement, but I don't quite see the disadvantage to the previous implementation. One can still scroll up and down, the selection just follows. Having a setting seems way over the top for something that I only see as *nice to have* :smile:

jcjgraf:
Makes sense. Feel free to merge this in this case :+1:  But maybe we want to open an issue concerning the horizontal scrolling.

karlch:
I took a shot at this as it didn't seem as complicated as I somehow thought, so at least for the touchpads it should work as expected, let me know if it also works for yours :blush: I would then only add the corresponding mouse keybindings and merge.

jcjgraf:
Works flawless for me on the touchpad. Thanks!

Issue #382, opened 2021-05-03
Improve Documentation


jcjgraf:
Some points how the documentation could be improved:

- [ ] [Settings overview](https://karlch.github.io/vimiv-qt/documentation/configuration/settings.html#overview-of-all-settings)
    - [ ] Indicate the domain of the setting options.
        - Some options may be more obvious (e.g. boolean settings) while others require analysing the source to figure out allowed values.
    - [ ] Indicate the default value.

- [ ] Document command line arguments
    - Briefly mention and describe all command line option. In addition a section with a few examples may be useful, especially after merging #381

karlch:
I agree that more documentation is always nice!

For both I would like to make sure that most of it stays fully automated, i.e. no need to manually update the documentation when we add another setting / command line flag. Additional examples and explanations are of course not part of this automation.

The question then is, how detailed we want to be. Would it be enough if we expand the settings table with two more columns, namely `type` (e.g. `bool`) and `default` (e.g. `true`)? We could possibly have a brief part before / after the table explaining any non-intuitive types, like the thumbnail size, without going into more details for every setting.

For the commandline arguments, is the information provided by `vimiv --help` enough except for the further examples?

jcjgraf:
Keeping it simple by indicating the `type` including possible min/max values and `default` is sufficient I think. Using the handy `scripts/src2rst.py` and your nicely structured `vimiv/api/settings.py` it should doable I guess. If you do not mind I would look into this some time after #325 is done.

Concerning the CLI arguments. I think the information provided by the `--help` is enough if accompanied with a bunch of examples.

PR #385, opened 2021-05-03
Add commandline flag to pass arguments directly to qt

### LINKED ISSUE #384: Setting WM_CLASS or WM_CLASS_INSTANCE
Hi ! Me again !

I use a tilling window manager (`bspwm`) and I would like to be able to set a rule such as:
```sh
bspc rule -a vimiv:float state=floating
```

To automatically set some `vimiv` instances to floating, typically when launching `vimiv` from a script to kind of emulate a `rofi` popup.

I saw that I can change the title through:

`vimiv --set title.fallback float`

But with `bspwm` (as far as I can tell) it isn't too easy to set a rule based on a title and it feels a bit hacky.

Ideally I would like to be able to change the WM_CLASS or WM_CLASS_INSTANCE of `vimiv`.

`sxiv` does allow this through the `-N` argument.

I quickly had a look at how to go about doing this with `pyqt5`, but I couldn't find a way acheive it but I have little experience with `pyqt5`.

karlch:
This is a possible solution to #384 which can now be done by starting vimiv with `--qt-args '--name WM_CLASS_INSTANCE'`. The implementation here is nowhere near as complete as the one of qutebrowser, but worked fine for the simple use-cases I could think of. With this, we could even deprecate / remove the `--geometry` flag as this is now possible using `--qt-args '--geometry WIDTHxHEIGHT'`. All of this would certainly benefit from some additional examples as described in #382.

@loiccoyle maybe you could give this a try as it is based on your idea?

Fixes #384.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -58,6 +58,9 @@ Added:
 * The ``-o TEXT`` flag to print ``TEXT`` to standard output before exit. Wildcards are
   expanded before printing, allowing to for example print marked files using ``-o %m``.
 * The ``-i`` flag to read paths to open from standard input instead.
+* The ``--qt-args`` flag to pass arguments directly to qt. When passing multiple
+  arguments, they must be quoted accordingly, e.g.
+  ``vimiv --qt-args '--name floating'``.
 
 Changed:
 ^^^^^^^^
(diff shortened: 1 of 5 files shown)


loiccoyle:
Wow that quick! I tried the WM_CLASS_INSTANCE floating window usecase I described in #384 and it worked flawlessly 🎉

karlch:
Awesome, thanks for checking! I will just merge this then.

For future reference:
I played around using the qt arguments directly for the geometry but found the current behaviour a bit nicer and easier to keep consistent.

PR #387, opened 2021-05-05
Setting to set the desired file ordering

### LINKED ISSUE #75: Different image sorting than by name

jcjgraf:
Adds settings to specify the desired ordering of the images and the desired order of folders.

Todo:
- [x] Seperate options for folder and images
- [x] `Size` for directory not working
- [x] Alias to conveniently change the ordering; @karlch do you like them or shall I get rid of them again?
- [ ] Implement other types of ordering. @all any ideas?
- [ ] Test
    - [x] Unit test natural sort
    - [ ] files.order using mock
    - [x] OrderSetting
    - [ ] end2end 
- [ ] Changelog

Partially based on #76. Some code was taken directly from there

resolves #75

### DIFF
--- a/tests/unit/utils/test_files.py
+++ b/tests/unit/utils/test_files.py
@@ -87,12 +87,6 @@ def test_listdir_wrapper_returns_abspath(mocker):
     assert files.listdir("directory") == expected
 
 
-def test_listdir_wrapper_sort(mocker):
-    mocker.patch("os.listdir", return_value=["b.txt", "a.txt"])
-    mocker.patch("os.path.abspath", return_value="")
-    assert files.listdir("directory") == ["a.txt", "b.txt"]
-
-
 def test_listdir_wrapper_remove_hidden(mocker):
     mocker.patch("os.listdir", return_value=[".dotfile.txt", "a.txt"])
     mocker.patch("os.path.abspath", return_value="")
@@ -120,6 +114,13 @@ def test_images_supported(mocker):
     images, directories = files.supported(["a", "b"])
     assert images == ["a", "b"]
     assert not directories
+
+
+def test_order():
+    assert files.order((["b.txt", "a.txt"], ["b", "a"])) == (
+        ["a.txt", "b.txt"],
+        ["a", "b"],
+    )
 
 
 def test_tar_gz_not_an_image(tmp_path):
--- a/tests/unit/utils/test_files.py
+++ b/tests/unit/utils/test_files.py
@@ -116,11 +116,25 @@ def test_images_supported(mocker):
     assert not directories
 
 
-def test_order():
-    assert files.order(["b.txt", "a.txt"], ["b", "a"]) == (
-        ["a.txt", "b.txt"],
-        ["a", "b"],
-    )
+@pytest.mark.parametrize(
+    "input_img, input_dir, sorted_img, sorted_dir",
+    (
+        (
+            ["a.j", "c.j", "b.j"],
+            ["a", "c", "b"],
+            ["a.j", "b.j", "c.j"],
+            ["a", "b", "c"],
+        ),
+        (
+            ["a2.j", "a3.j", "a1.j"],
+            ["a2", "a3", "a1"],
+            ["a1.j", "a2.j", "a3.j"],
+            ["a1", "a2", "a3"],
+        ),
+    ),
+)
+def test_order(input_img, input_dir, sorted_img, sorted_dir):
+    assert files.order(input_img, input_dir) == (sorted_img, sorted_dir)
 
 
 def test_tar_gz_not_an_image(tmp_path):
--- a/tests/unit/utils/test_files.py
+++ b/tests/unit/utils/test_files.py
@@ -117,7 +117,7 @@ def test_images_supported(mocker):
 
 
 def test_order():
-    assert files.order((["b.txt", "a.txt"], ["b", "a"])) == (
+    assert files.order(["b.txt", "a.txt"], ["b", "a"]) == (
         ["a.txt", "b.txt"],
         ["a", "b"],
     )
--- a/vimiv/api/settings.py
+++ b/vimiv/api/settings.py
@@ -13,7 +13,8 @@ Module attributes:
 import abc
 import contextlib
 import enum
-from typing import Any, Dict, ItemsView, List
+from typing import Any, Dict, ItemsView, List, Callable, Tuple
+import os
 
 from PyQt5.QtCore import QObject, pyqtSignal
 
@@ -315,6 +316,37 @@ class StrSetting(Setting):
         return "String"
 
 
+class ImageOrderSetting(Setting):
+    """Stores a image ordering setting."""
+
+    typ = str
+
+    ORDER_TYPES = {
+        "name": (os.path.basename, False),
+        "name-desc": (os.path.basename, True),
+        "modify": (os.path.getmtime, False),
+        "modify-desc": (os.path.getmtime, True),
+        "size": (os.path.getsize, False),
+        "size-desc": (os.path.getsize, True),
+    }
+
+    def convert(self, value: str) -> str:
+        if value not in self.ORDER_TYPES:
+            raise ValueError(f"Option must be one of {', '.join(self.ORDER_TYPES)}")
+
+        return value
+
+    def get_para(self) -> Tuple[Callable[..., Any], bool]:
+        """Returns the ordering parameters according to the current set value."""
+        try:
+            return self.ORDER_TYPES[self.value]
+        except KeyError:
+            raise ValueError(f"Option must be one of {', '.join(self.ORDER_TYPES)}")
+
+    def __str__(self) -> str:
+        return "ImageOrder"
+
+
 # Initialize all settings
 
 monitor_fs = BoolSetting(
@@ -333,6 +365,7 @@ style = StrSetting("style", "default", hidden=True)
 read_only = BoolSetting(
     "read_only", False, desc="Disable any commands that are able to edit files on disk"
 )
+image_order = ImageOrderSetting("image_order", "name", desc="Set ordering.",)
 
 
 class command:  # pylint: disable=invalid-name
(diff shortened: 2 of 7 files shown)


karlch:
Thanks a lot, I like where this is going! :blush: 

I am not quite sure about the aliases. On the plus-side, they are discoverable when typing `:order`. The disadvantage is that they don't have the nice completion that `:set ...` offers which makes them nearly useless for people that just "discovered" them without looking at the documentation.

Let me know when you are ready for a more detailed review.

jcjgraf:
The discoverability was indeed the main reason why I have added the aliases. But as you say, without the completion they are anyway useless. I will remove them again...

Function-vise I am done. So I guess you can start reviewing that.

Concerning the test, I think I need some advice from you. So far I have added the some basic settings tests in `test_settings` and fixed a testcase in `test_files` (which expected `files.listdir` to list files ordered). Now, should every single ordering method be testes? If so, in what file? Also, would it be a good Idea to add an end2end test to test if the order in the library really changes when changing this option? What else should I add testcases for?

I have also thought about renaming the ordering as follows. Do you have any opinion or suggestions?

| Current           | New                    |
| :---              | :---                   |
| name              | alphabetical           |
| name-desc         | alphabetical-desc      |
| name-natural      | natural                |
| name-natural-desc | natural-desc           |
| modify            | last-modified          |
| modify-desc       | most-recently-modified |
| size              | smallest               |
| size-desc         | largest                |

tests/unit/utils/test_files.py
```diff
@@ -122,6 +116,13 @@ def test_images_supported(mocker):
     assert not directories
 
 
+def test_order():
```
karlch, line 119: We may want to mock the api setting here

vimiv/api/working_directory.py
```diff
@@ -224,7 +226,7 @@ def _get_content(self, directory: str) -> Tuple[List[str], List[str]]:
         """
         show_hidden = settings.library.show_hidden.value
         paths = files.listdir(directory, show_hidden=show_hidden)
-        return files.supported(paths)
+        return files.order(files.supported(paths))
```
karlch, line 229: This is an example where we would need to unpack, i.e. `files.order(*files.supported(paths))` (see later comment in files.py).

We may want a convenience function that does
```python
files.order(*files.supported(files.listdir(directory)))
```

vimiv/api/settings.py
```diff
@@ -333,6 +409,12 @@ def __str__(self) -> str:
 read_only = BoolSetting(
     "read_only", False, desc="Disable any commands that are able to edit files on disk"
 )
+image_order = ImageOrderSetting(
```
karlch, line 412: We would then always create a `OrderSetting` and pass the additional options for images / directories here.

karlch (commented):
Thanks again, I really like where this is going and am very fond of the new names for the ordering types. I left a few minor comments concerning the implementation.

Concerning the tests, I am fine with not testing every single ordering function that is trivial, however the natural sorting would be nice to test in the `test_settings` module as it is a class method.

One end2end test checking if the order changes in the library when the setting is changed would be great :blush:

vimiv/api/settings.py
```diff
@@ -315,6 +317,80 @@ def __str__(self) -> str:
...
+        def convert(t: str) -> Union[str, int]:
+            return int(t) if t.isdigit() else t
+
+        return [convert(c) for c in re.split(r"(\d+)", text)]
+
+    def __str__(self) -> str:
+        return "Order"
+
+
+class ImageOrderSetting(OrderSetting):
```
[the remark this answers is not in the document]
jcjgraf: I have merged the classes in 78ad67e7879ea4250b85e8b3e62b4c401b39404a

However, I do not manage to reference the static method `_natural_sort` from the class variable `ORDER_TYPES`. Can you hint me to the right syntax?

vimiv/api/settings.py
jcjgraf: Implemented in 78ad67e7879ea4250b85e8b3e62b4c401b39404a.

vimiv/api/settings.py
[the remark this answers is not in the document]
karlch: Ah, good point. In the current way you will not be able to access the `_natural_sort` method as it is defined later than the dictionary. I see two options:
- Just define this as a regular function, could even go into `utils` as far as I am concerned.
- Instead of defining the dictionary as a class attribute, you could have a function that returns this dictionary, something along:
```python
@classmethod
def _default_order_types(cls):
    return {
        ...
        "natural": cls._natural_sort
    }
```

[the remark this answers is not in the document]
jcjgraf: Thanks for the info, I have moved it to `utils`.

vimiv/utils/files.py
```diff
@@ -55,6 +57,26 @@ def supported(paths: Iterable[str]) -> Tuple[List[str], List[str]]:
     return images, directories
 
 
+def order(files: Tuple[List[str], List[str]]) -> Tuple[List[str], List[str]]:
```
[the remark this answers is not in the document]
jcjgraf: Added in 14621f6eb7e14de966618e993ceec31c1438c127.

vimiv/api/working_directory.py
jcjgraf: I have changed the unpacking in 14621f6eb7e14de966618e993ceec31c1438c127 and adopted the calling of `files.order` adequately.

I do not fully understand the purpose of this convenience function, since this is in essence exactly what `WorkingDirectoryHandler._get_content()` does. Or do you mean we should *move* this funcionality to `utils.files`?

vimiv/api/settings.py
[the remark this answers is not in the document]
karlch: `__init__` is great, in principle one could also sort non-file objects with this :blush:

vimiv/api/working_directory.py
karlch: You are right, looks good to me :+1:

vimiv/api/settings.py
```diff
@@ -315,8 +316,51 @@ def __str__(self) -> str:
...
+
+    def get_para(self) -> Tuple[Callable[..., Any], bool]:
+        """Returns the ordering parameters according to the current set value."""
+        try:
+            return self.order_types[self.value]
+        except KeyError:
+            raise ValueError(f"Option must be one of {', '.join(self.ORDER_TYPES)}")
 
+    def suggestions(self) -> List[str]:
+        return [str(value) for value in self.order_types]
```
karlch, line 357: Mini nit-pick, `list(self.order_types)` should suffice here, all keys are strings anyway, no?

karlch:
This is now almost complete, right? Just some minor work on the testing, add to changelog, and then we are ready to merge? :blush:

jcjgraf:
Thanks for your comments, it will look into them. Yes, this PR is pretty much complete. I just have to create an E2E test to test the general behaviour and mock the `test_order`, as indicated by you first comment. Since I am still not that familiar with PyTest (or testing in general) it may take some time till I am done with that :upside_down_face:

karlch:
Sounds good, as always take all the time you need :blush:

karlch:
I had a brief look at this, seems almost ready to go, would you mind if I merge to a branch and try to finalise this while you work on #467 over the week-end?

jcjgraf:
Indeed, functionality-vise everything should be done. However, I have not looked at the code myself for over 1 year. 🙈 You can merge and finalize it, or let me rebase and go over it again. However you like 😊

vimiv/api/settings.py
jcjgraf: Indeed true. That is much cleaner

karlch:
I'm happy to take a shot, I already rebased and looked into the code, should be rather straightforward as it is looking good :blush:

`inv` sound good, I would actually even write it out, i.e. `inverse` as we have completion anyway, so shortening shouldn't be needed.

jcjgraf:
Sounds good, happy coding session :blush:

And feel free to squash all my commits as the history is rather dirty :see_no_evil:

vimiv/utils/files.py
```diff
@@ -26,13 +28,18 @@ def listdir(directory: str, show_hidden: bool = False) -> List[str]:
...
-        os.path.join(directory, path)
-        for path in os.listdir(directory)
-        if show_hidden or not path.startswith(".")
+        [
+            os.path.join(directory, path)
+            for path in os.listdir(directory)
+            if show_hidden or not path.startswith(".")
+        ],
+        key=order_function,
+        reverse=reverse,
```
karlch, line 42: I am currently looking over this, and not quite sure it makes sense to sort by the api setting here. Listdir is essentially used in three places.

1. To get the files in the library. These are later sorted according to directory / image setting anyway.
2. To get the files for the path completion. These should be sorted explicitly, to ensure both image_order and directory_order are respected.
3. To get the files for trash completion. We could use the image_order here, as we don't support deleting directories, but I believe explicitly using it in the completion makes more sense as well.

Thus I would remove the sorting from this all-together. Let me know if I missed something :blush:

tests/unit/utils/test_files.py
karlch: no longer needed in the new implementation, as sorting is done in the setting directly

vimiv/utils/files.py
jcjgraf: I guess this has been resolved by 0e4a0db59d977a40638230e1a06e867034ee989f

jcjgraf: >just wanted to make sure I didn't miss a reason

That is what I though. But do not rely on me too much. I only learned thanks to this review and your commits where else the sorted file paths are also used :laughing:

jcjgraf:
Closing, as #563 has been merged.

PR #390, opened 2021-05-11
Describe Command Line Arguments

jcjgraf:
Adds a section containing all available command line arguments to the documentation

Todos:
- [x] Fix `--` displayed as `-`
- [x] Add examples
- [x] Improve phrasing
- [x] (Re)Think about placement in documentation

Implements the second task from #382


karlch:
Thanks a lot for starting to work on this! Just let me know when you are ready for a quick review so we can start the discussion :blush:

jcjgraf:
Content-vies it contains everything. So feel free to have a close look at it. 
The last two todos are anyways rather for you; if you do not like the phrasing or spot any typos let me know. Also, I have created a dedicated section for the commandline arguments. But maybe you prefers to have it integrated in a pre-existing section like the `Getting Started` section.

docs/documentation/cl_options/index.rst
```diff
@@ -0,0 +1,40 @@
+Command Line Arguments
+======================
+
+What starting vimiv form the command line you have the ability to pass a number of
```
karlch, line 4: *When* starting vimiv *from* ...

docs/documentation/cl_options/index.rst
```diff
@@ -0,0 +1,40 @@
...
+different argument to vimiv. The general calling structure is depicted below:
```
karlch, line 5: I would leave out the "depicted below", just
"The general calling structure is:"

docs/documentation/cl_options/index.rst
```diff
@@ -0,0 +1,40 @@
...
+
+.. include:: synopsis.rstsrc
+
```
karlch, line 8: I would put the examples in front of the detailed list. As a reader the examples are more interesting, when looking for a reference you will most likely click on it directly and not "discover" it.

docs/documentation/cl_options/index.rst
```diff
@@ -0,0 +1,40 @@
...
+The following is an exhaustive list of all available arguments:
+
+.. include:: options.rstsrc
+
+Examples
+--------
+
+In the following we present a few use cases of command line arguments.
+
```
karlch, line 20: I find the formatting here a bit weird. How about formatting each of the points like this:
```
* Start in thumbnail view instead of image view::

   vimiv * --command "enter thumbnail" --command "enter library"
```
Note that both the number of spaces and the empty line are syntax, restructured text can be a pain :laughing:

docs/documentation/cl_options/index.rst
```diff
@@ -0,0 +1,40 @@
...
+* Start in thumbnail view instead of image view
+    ``vimiv * --command "enter thumbnail" --command "enter library"``
```
karlch, line 22: I would remove the `--command "enter library"` here as the text says "start in thumbnail view". You could add a second point stating "start in library view with the thumbnail grid displayed" for the combination if you like.

docs/documentation/cl_options/index.rst
```diff
@@ -0,0 +1,40 @@
...
+
+* Start in read-only mode. This prevents accidental modification (renaming, moving, editing etc.) of any images
+    ``vimiv --set read_only true``
+
+* Change WM_CLASS_INSTANCE to identify a vimiv instance
+    ``vimiv --qt-args '--name myVimivInstance'``
```
karlch, line 28: Nit-picking, but you use double-quotes everywhere else, so I would also use them here :blush:

docs/documentation/cl_options/index.rst
```diff
@@ -0,0 +1,40 @@
...
+
+* Print the last selected image to STDOUT when quitting
+    ``vimiv --output "%"``
+
+* Use vimiv as *Rofi for Images* to make a selection from candidate image
```
karlch, line 33: from candidate *images*

docs/documentation/cl_options/index.rst
```diff
@@ -0,0 +1,40 @@
...
+    ``mySel=$(echo $myCand | vimiv --input --output "%m" --command "enter thumbnail")``
+
+* Print debug logs of your amazing plugin you are writing and of the `api._mark` module which does not behave as you are expecting
+    ``vimiv --debug myAmazingPlugin api._mark``
+
+* Collect data to report this weird bug you are experiencing
```
karlch, line 39: Curious what the use-case of this is compared to just using `~/.local/share/vimiv/vimiv.log` with `vimiv --log-level debug`

scripts/src2rst.py
```diff
@@ -91,22 +91,50 @@ def _gen_keybinding_rows(bindings):
     return header + rows
 
 
-def generate_commandline_options():
-    """Generate file including the command line options."""
+def generate_commandline_options(for_man=True):
```
karlch, line 94: I am not quite sure I like the structure here. Some general thoughts:
- We could call `generate_commandline_options` only once and the function deals with doing both, splitting off where needed.
- The implementation for the website seems like a bit of a work-around as it parses the string which was formatted for the man page. Maybe the formatting should be done individually, and only the actual information is grabbed in a shared way?
- Currently it seems like the identical synopsis file is written twice. We could avoid this by writing to `docs/synopsis.rstsrc` and including it using `.. include:: /synopsis.rstsrc`. But,
- As you point out the formatting of bold regular text is not ideal for the website, as it shows up in a weird way. I guess we could try out using code ticks instead of bold for there?

This will probably result in quite a rehaul of this part as it was designed only for the man page. What do you think?

karlch (commented):
Thanks a lot again! I left a few comments. The ones concerning the documentation should be straightforward and simple quick-fixes, the code may need some more re-structuring. As always I am happy to help and up for any discussion :blush:

scripts/src2rst.py
jcjgraf: The conversion from the manpage formatting to the list of tuples is indeed a bit hacked and not very nice. Are you aware of a method to get this data from the argument parser in some different format? Or what do you mean by your second point?

karlch: This is a code snippet I came up with in a debug script. Instead of reading through the parser and formatting the arguments, we just grab everything we need and store it into a list of named tuples. We then have the "raw" information which could be formatted accordingly for the man page and website. There may be some parts of shared formatting that could also go into the `ParserArgument` class, e.g. the formatting of what is called title here.

<details><summary>Quick script to get information from the argparser object</summary><p>

```python
import typing

from vimiv import parser


class ParserArgument(typing.NamedTuple):
    """Storage class for a single command line argument."""
    group: str
    longname: str
    shortname: str
    metavar: typing.Any
    description: str

    @property
    def is_positional(self):
        return self.longname is self.shortname is None

    def title(self, formatter: str = "**"):
        ...


def get_arguments(argparser) -> typing.List[ParserArgument]:
    """Retrieve all arguments from the passed argparser."""
    def argument_from_action(group, action):
        if len(action.option_strings) == 2:
            longname, shortname = action.option_strings
        elif len(action.option_strings) == 1:
            longname = action.option_strings[0]
            shortname = None
        else:
            longname = shortname = None
        return ParserArgument(
            group=group.title,
            longname=longname,
            shortname=shortname,
            metavar=action.metavar,
            description=action.help,
        )

    return [
        argument_from_action(group, action)
        for group in argparser._action_groups
        for action in group._group_actions
    ]


def main():
    argparser = parser.get_argparser()
    arguments = get_arguments(argparser)
    # We should now have all the "raw" information which can then be formatted
    # accordingly for the different options
    # Possibly shared code could go into the ParserArgument class with arguments, e.g.
    # formatting the title
    for elem in arguments:
        print(elem)


if __name__ == "__main__":
    main()
```

</p></details>

docs/documentation/cl_options/index.rst
jcjgraf: rebased and fixed in 5aa4b56a13f4f7aa76d208b9a10b6e891e525ad6

docs/documentation/cl_options/index.rst
jcjgraf: rebased and fixed in e6a562414448933601ee2b9b8f0000c9fe266fc2

docs/documentation/cl_options/index.rst
jcjgraf: rebased and fixed in 5aa4b56a13f4f7aa76d208b9a10b6e891e525ad6

docs/documentation/cl_options/index.rst
jcjgraf: I agree that the formatting could be better. Though, with your proposed solution double quotes look somewhat odd (probably depending on ones font) and more importantly `--` gets displayed as `-`. In e6a562414448933601ee2b9b8f0000c9fe266fc2 I have mixed yours any my solution. I think this way it looks quite decent. Let me know if you have a better solution.

docs/documentation/cl_options/index.rst
jcjgraf: I have improved the wording as you have suggested in e6a562414448933601ee2b9b8f0000c9fe266fc2.

docs/documentation/cl_options/index.rst
jcjgraf: Good catch. That's caused by me copy/pasting examples from different issues :upside_down_face: 

fixed in e6a562414448933601ee2b9b8f0000c9fe266fc2

docs/documentation/cl_options/index.rst
jcjgraf: rebased and fixed in e6a562414448933601ee2b9b8f0000c9fe266fc2

docs/documentation/cl_options/index.rst
jcjgraf: Ok well, your right. I have removed it in e6a562414448933601ee2b9b8f0000c9fe266fc2

docs/documentation/cl_options/index.rst
karlch: This is how the two versions, the one I proposed and the one in your commit, look for me:

scrot-2021-05-18_21:46:38

As you can see there is no issue with the double quotes and the `--` characters. Maybe you omitted the two colons? To my understanding the syntax with the two colons is more or less identical to using a `.. code-block`, but I haven't been able to properly indent using an explicit code block :smiley: 

In any case, both options are fine for me, with a slight preference for the gray block.

jcjgraf: I have indeed missed the two colons :upside_down_face:  It looks much better this was, I have changes it in 5983290df343667c7835c5df4c4db94b0220a9a8

karlch: Thanks! Restructured text syntax can be a real pain...

scripts/src2rst.py
jcjgraf: I have restructured the generator code based on you code snippet in 7226827ec31178e7978f66a21c81eb2e46a1a115. I hope it kind of reflects your idea. It definitely gets rid of the hacked man page to documentation fomat conversion. Let me know if you spot anything to improve :upside_down_face: 

I still do not really like the formatting of the synopsis in the documentation. I would prefers e.g. grey code box, but `::code-block:` does unfortunately not add line breaks.

karlch: I will need a bit more time to read through the actual code, but it does look a lot better on a first glance. We may want to split this off into its own file though and import it in `src2rst` as it is getting rather involved.

Concerning the synopsis, we could let python do the wrapping for us using the stdlib `textwrap` module, e.g.
```python
    if man_format:
        return synopsis
    return "\n   ".join(textwrap.wrap(synopsis, width=79))
```

jcjgraf: Thanks for the hint with `textwrap`. I have added that. It looks much better now.

I did not yet split the code. I suggest that you review the code once you have time for that and tell me if you spot anything to improve/fix and if you prefer to move the cli option generation part into a separate file.

jcjgraf:
Concedrning the functionality and the content, this PR is done. You have mentioned, that we may want to split the code into multiple files. Do you have anything specific splitting in mind?

I can think of:
1) Move only `ParserArgument`
2) Move in addition `_get_arguments`, `_generate_synopsis`, `_generate_command`
3) Move in addition `generate_commandline_options`

What you think?

karlch:
I am not sure concerning the splitting anymore, and really still have to review this properly, sorry for the delays. One thing I noticed though is that the man page looks different now, the `SYNOPSIS` part now includes some of the `*` characters which should only be used for formatting. I haven't figured out why this is the case though..

jcjgraf:
Are you talking about `docs/manpage/synopsis.srtsrc`? Is there way to have a look at the final man page, as `vimiv --help` seems to be still showing some different version?

Just looking at the `synopsis.srtsrc`, everything looks fine to me:

Initial:
```
**vimiv** [**PATH**] [**-h**] [**--help**] [**-f**] [**--fullscreen**] [**-v**] [**--version**] [**-g** *WIDTHxHEIGHT*] [**--geometry** *WIDTHxHEIGHT*] [**--temp-basedir**] [**--config** *FILE*] [**--keyfile** *FILE*] [**-s** *OPTION* *VALUE*] [**--set** *OPTION* *VALUE*] [**--log-level** *LEVEL*] [**--command** *COMMAND*] [**-b** *DIRECTORY*] [**--basedir** *DIRECTORY*] [**-o**] [**--output**] [**-i**] [**--input**] [**--debug** *MODULE*] [**--qt-args** *ARGS*]
```

This PR:
```
**vimiv** [**PATH**] [**--help**|**-h**] [**--fullscreen**|**-f**] [**--version**|**-v**] [**--geometry** *WIDTHxHEIGHT*|**-g** *WIDTHxHEIGHT*] [**--temp-basedir**] [**--config** *FILE*] [**--keyfile** *FILE*] [**--set** *OPTION* *VALUE*|**-s** *OPTION* *VALUE*] [**--log-level** *LEVEL*] [**--command** *COMMAND*] [**--basedir** *DIRECTORY*|**-b** *DIRECTORY*] [**--output**|**-o**] [**--input**|**-i**] [**--debug** *MODULE*] [**--qt-args** *ARGS*]
```

karlch:
You can build the man page via
```
tox -e man
```
and then open it directly
```
man misc/vimiv.1
```

With this PR I get a final look like this:
scrot-2021-08-08_18:55:41
which is somewhat off, but I am not sure why yet.

jcjgraf:
Soon is new year and I sill have quite a few dangling PRs. Therefore, I thought it is a good time to start finish them :blush:

I have gone through all changes of this PR again, rebased it to the current master and fixed the formatting error you have noticed.

There are a few remaining points to discuss:
- You have mentioned that we may want to split `src2rst.py` up. Personally, I do not think that is actually that bad. If it is getting out-of-hands when continuing with #382 we can still split it later on. But it is up to you to decide.
- It would be nice to indicate all possible values for `--qt-args`. The problem is that I do not know the allowed args (nor did I find them in the PyQt docs). It would be nice to have them generated automaicalls, or if this is not possible, add a link to some external documentation or something. But we can still do that later on I case you are not aware of an simple solution.

Besides the mentioned points, I am done with this PR. Let me know if there is anything to improve or fix.

karlch:
Thanks for your additional effort!

Concerning your two points:
- I agree, we can keep it for now, but probably need re-formatting sooner or later.
- I am also not sure where one would find all possible values, but probably many of them would not be useful for us anyway. One example would be the `-reverse` flag which causes vimiv to look horribly broken :laughing: So I would ignore this for now and stick to documenting useful (working) options.

I am a bit confused with the state of this PR though, when I try to build anything I get an AttributeError:
```shell
man run-test: commands[0] | /home/christian/Coding/VimivQt/scripts/src2rst.py
Traceback (most recent call last):
  File "/home/christian/Coding/VimivQt/scripts/src2rst.py", line 182, in <module>
    def _get_arguments(argparser: parser.argparser.ArgumentParser) -> List[ParserArgument]:
AttributeError: module 'vimiv.parser' has no attribute 'argparser'. Did you mean: 'argparse'?
```
and when I fix this simple error I get more havoc
```shell
man run-test: commands[0] | /home/christian/Coding/VimivQt/scripts/src2rst.py
generating statusbar modules...
generating commands...
generating settings...
generating keybindings...
Traceback (most recent call last):
  File "/home/christian/Coding/VimivQt/scripts/src2rst.py", line 333, in <module>
    generate_commandline_options()
  File "/home/christian/Coding/VimivQt/scripts/src2rst.py", line 105, in generate_commandline_options
    f.write(_generate_synopsis_man(arguments))
  File "/home/christian/Coding/VimivQt/scripts/src2rst.py", line 228, in _generate_synopsis_man
    command = "\ \|\ ".join(arg.get_name_metavar())
TypeError: ParserArgument.get_name_metavar() missing 2 required positional arguments: 'name_formatter' and 'metavar_formatter'
```

Am I missing something?

jcjgraf:
I am sorry, I must have lost a commit while rebasing/cleanup. Now
everything (should) work!

jcjgraf:
quickly rebased to solve merge conflicts

karlch:
Thanks again for working on this! When looking through,  I thought there were some small parts of code that could be simplified (were you on Haskell drugs at the time? :wink: :yum: ). Thus, I quickly did the changes, and also added the settings part in #657.

Hope this is okay with you! :blush: 

If you like, you can take a quick look at #657 and give your okay :blush:

PR #394, opened 2021-05-21
Fix statusbar inconsistencies

jcjgraf:
There were some inconsistencies between the base modules and some thumbnail-* modules. Namely, `basename` contained the file extension while it was striped for `thumbnail-name`, and `index` was zero padded while `thumbnail-index` was not.

The first point was fixed by distinguishing between `name` (name w/o extension), `basename` (name w extension) and `extension` (i.e. adding seperate modules for each of them).

The default statusbar module for `left_thumbnail` was changed from `thumbnail-name` to `thumbnail-basename` to be consistent with `left_image`.

@karlch feel free to merge if you agree

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -64,6 +64,8 @@ Added:
   arguments, they must be quoted accordingly, e.g.
   ``vimiv --qt-args '--name floating'``.
   Thanks `@loiccoyle`_ for the discussion!
+* Statusbar modules ``name``, ``thumbnail-basename``, ``extension`` and
+  ``thumbnail-extension``.
 
 Changed:
 ^^^^^^^^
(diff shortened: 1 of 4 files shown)


karlch:
Thanks for catching these! While I am not 100 % sure I like the naming, i.e. I find it hard to relate `name` and `basename` to "basename without extension" and "basename with extension", I cannot really think of anything better and like the fact that this is more-or-less backwards compatible in the sense that all statusbar modules that existed before still exist and no new errors are raised.

jcjgraf:
I agree with your points; Backwards compatibility was the main reason why I did not rename the them to something more meaningful. If we come up with something more useful we could replace then and add `basename` as an alias. Once 1.0.0 is released we could get rid of the alias again.

Issue #374, opened 2021-04-26
[BUG] Uncaught exception when searching


BachoSeven:
This happens in library mode, when I press `/` to open the search prompt and type any letter. The only notable thing inside the folder in which this happens is the the `4k` folder contains a bunch of symlinks, and two files starting with the `!` character. Tell me if you need more info.

OS: ArchLinux
vimiv-qt version: `r2162.b4b7772f`

``` python
[18:49:37] ERROR    <vimiv>              Uncaught exception! Exiting gracefully and printing stack...
Traceback (most recent call last):
  File "/usr/lib/python3.9/site-packages/vimiv/gui/commandline.py", line 137, in _incremental_search
    search.search(text, self.mode, reverse=prefix == "?", incremental=True)
  File "/usr/lib/python3.9/site-packages/vimiv/commands/search.py", line 71, in __call__
    self._run(text, mode, count, reverse, incremental)
  File "/usr/lib/python3.9/site-packages/vimiv/commands/search.py", line 89, in _run
    current_index = paths.index(api.current_path(mode))
ValueError: '/mnt/01D5707D40B081F0/Users/ascol/Pictures/papes/4k' is not in list
```

karlch:
Thanks for the report!

Unfortunately I have not been able to reproduce the crash just by adding a few symlinks or files starting with the `!` character to a clean directory and starting a search. Would it be possible for you to reproduce this from an empty directory and record the steps needed to create this directory which causes the crash?

BachoSeven:
Steps: 
- create a directory `mkdir test`
- make a symlink of it: `ln -s test testln`
- create a folder named `4k` inside of it: `cd testln && mkdir 4k && cd ..`
- `vimiv`, then navigate inside `testln` and then press `/` followed by any character
- > Crashes

karlch:
Thanks a lot for this detailed report, I can now reproduce this fine and believe I undersand the issue.

When vimiv enters `testln` we truly enter `test` and this is also what `os.getcwd` / `os.path.abspath` and friends will return. The filelist vimiv populates is however a simple line along `os.path.join("testln", "4k")`. This ultimatley results in the current path `test/4k` not existing in the filelist which only knows `testln/4k`.

The quick solution I can see is replacing `os.path.abspath` with `os.path.realpath` when appropriate. Changing only the `listdir` function in `vimiv.utils.files` to
```python
    directory = os.path.realpath(os.path.expanduser(directory))
    return sorted(
[...]

PR #397, opened 2021-06-03
Fix showing both real and symlink image from different directories

karlch:
The new usage of realpath introduced in 0cc236327662dbf18c5f1d82cee287941d4e1350 also introduced a new bug. The image filelist would load both the symlink and the original image, even if they are not in the same directory. This PR attempts to save this issue while also fixing the initial issue that the mentioned commit tried to address.

Related to #374

### DIFF
--- a/tests/end2end/features/misc/symlink.feature
+++ b/tests/end2end/features/misc/symlink.feature
@@ -1,14 +1,18 @@
 Feature: Handling symbolic links
 
-    Background:
-        Given I open the symlink test directory
-
     Scenario: Open real path when opening symlink
+        Given I open the symlink test directory
         When I run open lnstem
         Then the working directory should be stem
 
     Scenario: Crash when searching in symlinked directory
+        Given I open the symlink test directory
         When I run open lnstem
         And I run search
         And I press 'k'
         Then no crash should happen
+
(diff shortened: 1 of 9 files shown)


jcjgraf:
I can confirm that the newly introduced bug is fixed.

jcjgraf:
I could not reproduce the first problem described in #374, but I am not sure about the initial one. @BachoSeven could you maybe give it a try?

BachoSeven:
Hei, I tried to reproduce what I mentioned(i.e. [this](https://github.com/karlch/vimiv-qt/issues/374#issuecomment-827932686)), and it indeed does seem fixed, meaning entering that directory within the symlink(which itself then contains symlinks) (with vimiv from this branch) doesn't crash anymore.

BachoSeven:
If this is merged, it fixes  #374.

jcjgraf:
>  I hoped this would do the github thingie where it automagically closes the issue once the PR gets merged :'D

This only works if added to the PR description or to a commit message in this PR I guess

karlch:
Wow, been a while, shame on me but time to merge :smile:

jcjgraf:
Awesome, thanks for fixing and merging. This way I do not always have to checkout this branch when using Vimiv :blush:

Issue #74, opened 2019-06-06
Path completion for multiple paths


karlch:
As it is now possible to handle multiple paths in e.g. `:open`, the completion should start with new suggestions once a space is entered after the first complete path.

This requires an update to the `PathModel` class in `vimiv/completion/completionmodels.py`.  Currently the implementation is in the `on_text_changed` method but with multiple paths some refactoring and splitting into multiple methods may become necessary.

pspiagicw:
If nobody is working on this can I take this , I am a beginner and this looks easy .

jcjgraf:
Hey @pspiagicw, awesome that you are interested in contributing. Feel free to give it a try and open an PR for it. If you have any questions, or get stuck and need any guidance, do not hesitate to ask!

PR #413, opened 2021-08-01
Adds Multiple file completion

pspiagicw:
This PR adds multiple file completions mentioned in #74 .I have not looked into modifying/adding tests.

The commented lines in `_on_text_changed` method need discussing.

Any change is welcome!

### DIFF
--- a/vimiv/completion/completionmodels.py
+++ b/vimiv/completion/completionmodels.py
@@ -128,17 +128,8 @@ class PathModel(api.completion.BaseModel):
             total_directories.append(directory_list)
         
 
-        # Format data
-
-        # self.set_data([
-        #     self._create_row(os.path.join(directory, os.path.basename(path)))
-        #     for path in total_images + total_directories for directory in directories
-        #     ]
-        # )
         previous_completions = ['']
-        # print(directories ,total_images , total_directories)
         for directory , images , directory_list in zip(directories,total_images,total_directories):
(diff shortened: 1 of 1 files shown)


karlch:
Awesome, thanks a lot for your work and congrats on opening your first PR here :blush: 

I still see a few issues, but I am sure we can work around them together:
- We currently don't get completion immediately after typing `:open ` but have to add another character
- The usage of python's plain split means we broke completion for paths with spaces
- The performance seems extremely bad when in a directory with hundreds of files, I assume this scales with `n^x` where `n` is the number of files and `x` the number of paths for the completion. We should be able to get a simple `n` scaling though, as the first parts of the completion probably don't need to be re-calculated. What I mean is, when I have `:open x.jpg y.jpg ` the completion can keep `:open x.jpg and y.jpg ` and only add paths to the end, i.e. `:open x.jpg y.jpg z.jpg` or `:open x.jpg y.jpg a.jpg`, but we don't want `:open a.jpg y.jpg a.jpg` as the user already confirmed the first path should be `x.jpg`.

Does this make sense to you? If you like, I can also comment directly in the code with a review. Thanks again and happy hacking!

pspiagicw:
1 and 2 can be solved with a custom function instead of split function . 3 can also be solved using a custom function.

karlch:
I am not sure we need a custom function for the splitting, we can probably make use of the existing `api.completion.escape` / `api.completion.unescape` when working with paths and use `shlex.split` from the `shlex` module for splitting.

For 3. I think re-structuring the loop around 131-141 should solve the issue.

Let me know if any of this makes no sense :blush:

pspiagicw:
No Problem , I include these changes in the next commit.

Issue #420, opened 2021-09-04
Explicitly specify encoding when handling files


karlch:
As the newest pylint release pointed out with the `unspecified-encoding` warning, we do not specify the encoding when handling files. Starting from python 3.10 this will result in an [optional EncodingWarning](https://www.python.org/dev/peps/pep-0597/). We should therefore handle the encoding explicitly.

I currently see a few options:
- Use `encoding="utf-8"` everywhere. This is simple and compatible with the current behaviour on linux, the only operating system we currently test against and officially support.
- Use `encoding="utf-8"` where we are not sure about the content, and `encoding="ascii"` where we know ascii is enough. Requires thinking, more correct, not sure if worth it.
- Add a default encoding variable somewhere and use it whenever we don't want to specify explicitly. Almost certainly not worth it as the default will be `"utf-8"`.

What needs to be done:
- [x] Enable the `unspecified-encoding` check in `.pylintrc` by removing it from the `disable` part.
- [x] Fix every warning emitted by pylint.

jameslivulpi:
Hello, I will fix this

jameslivulpi:
pylint all warnings gone:
```
➜  vimiv-qt git:(iss420) pylint --rcfile .pylintrc * | grep "unspecified-encoding"
➜  vimiv-qt git:(iss420) 
```

karlch:
Thanks again, closing as fixed and merged to master :tada:

PR #422, opened 2021-09-05
adds encoding="utf-8" everywhere

jameslivulpi:
This takes care of  #420


karlch:
Awesome, that was really fast, thanks a lot!

Issue #363, opened 2021-04-14
Slow when opening pictures on slow file systems (e.g., remote file systems)


dschrempf:
When open one of many image in a directory of a slow file system (for example, remote file systems such as NFS, or SSHFS), `vimiv` takes a long time to start. I think this is because it caches many files in the directory, and not only the file to be opened. Is there a way to check if a file system is slow, and act accordingly? Many times, I only want to open a single file, and do not care about creating thumbnails for all other files in the directory.

Thanks for you help!

karlch:
Thanks for your interest in vimiv!

Unfortunately there is currently no such option, and I don't quite see how this would be implemented unless we add a simple setting. But of course it would be nicer to actually improve the performance in general, so we don't need any workarounds. In principle the thumbnails are created in parallel, so this should not delay startup, but all files in the current (the one of the first image) directory are checked if they are images or not which could be the cause of the delay.

I did some quick checks and found the performance to be pretty bad on slow file systems in general. The checking of all of the files becomes a significant part of the loading time (> 30 %) once there are around 100 images in the working directory. But even without this there are some parts that are surprisingly slow, profile attached for reference.

<details>
<summary>Startup profile</summary>
<p>

```
         27687 function calls (27276 primitive calls) in 3.331 seconds

   Ordered by: cumulative time
   List reduced from 916 to 15 due to restriction <15>

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.000    0.000    2.563    2.563 /home/christian/Coding/VimivQt/vimiv/startup.py:81(setup_post_app)
        1    0.000    0.000    1.167    1.167 /home/christian/Coding/VimivQt/vimiv/startup.py:113(init_paths)
        1    0.000    0.000    1.166    1.166 /home/christian/Coding/VimivQt/vimiv/api/__init__.py:57(open_paths)
        1    0.000    0.000    0.991    0.991 /home/christian/Coding/VimivQt/vimiv/api/working_directory.py:122(chdir)
        1    0.000    0.000    0.990    0.990 /home/christian/Coding/VimivQt/vimiv/api/working_directory.py:156(_load_directory)
        2    0.002    0.001    0.964    0.482 /home/christian/Coding/VimivQt/vimiv/utils/files.py:39(supported)
        1    0.000    0.000    0.958    0.958 /home/christian/Coding/VimivQt/vimiv/api/working_directory.py:218(_get_content)
[...]

PR #451, opened 2021-12-20
Improve startup time by reducing stat() calls

woefe:
On startup, multiple functions check properties of loaded files and
directories. Each of those functions like isdir(), isfile(), or
getsize() cause a stat() syscall. By calling stat() directly and reusing
the result we can improve startup time on remote and slow storage.

Related to #363 

Edit: This is relatively untested. Haven't checked what happens on reloads of the library etc.

### DIFF
--- a/vimiv/api/working_directory.py
+++ b/vimiv/api/working_directory.py
@@ -65,7 +65,7 @@ from PyQt5.QtCore import pyqtSignal, QFileSystemWatcher
 
 from vimiv.api import settings, signals, status
 from vimiv.utils import files, slot, log, throttled
-
+from vimiv.utils.files import SupportedMeta
 
 _logger = log.module_logger(__name__)
 
@@ -215,7 +215,7 @@ class WorkingDirectoryHandler(QFileSystemWatcher):
             self._directories = directories
             self.changed.emit(images, directories)
 
-    def _get_content(self, directory: str) -> Tuple[List[str], List[str]]:
+    def _get_content(self, directory: str) -> SupportedMeta:
         """Get supported content of directory.
 
         Returns:
@@ -224,7 +224,7 @@ class WorkingDirectoryHandler(QFileSystemWatcher):
         """
         show_hidden = settings.library.show_hidden.value
         paths = files.listdir(directory, show_hidden=show_hidden)
-        return files.supported(paths)
+        return files.supported_with_meta(paths)
 
 
 handler = cast(WorkingDirectoryHandler, None)
--- a/vimiv/api/working_directory.py
+++ b/vimiv/api/working_directory.py
@@ -156,8 +156,9 @@ class WorkingDirectoryHandler(QFileSystemWatcher):
     def _load_directory(self, directory: str) -> None:
         """Load supported files for new directory."""
         self._dir = directory
-        self._images, self._directories = self._get_content(directory)
-        self.loaded.emit(self._images, self._directories)
+        images, directories = self._get_content(directory)
+        self._images, self._directories = files.strip_meta((images, directories))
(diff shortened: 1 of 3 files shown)


woefe:
I did some profiling on a Samba-mounted directory with 341 photos. The situation before this PR: You can see that stat() calls can be quite expensive on remote storage and that there are 3 branches calling stat().
image

The situation after this PR. You can see below that there is now only one branch from the direct vimiv code to stat(). Calls to stat() are reduced by 2 * 341 = 682 calls. On my test setup the startup is now around 10s faster.
image

karlch:
Thanks a lot for your work on this, it looks very promising! I will try to take a closer look over the next few days.

Enjoy your Christmas time! :christmas_tree:

vimiv/api/working_directory.py
```diff
@@ -193,14 +194,19 @@ def _maybe_emit_image_changed(self) -> None:
...
         """Emit changed signals if the content in the directory has changed.
 
         Args:
-            images: Updated list of images in the working directory.
-            directories: Updated list of directories in the working directory.
+            images_meta: Updated list of images in the working directory.
+            directories_meta: Updated list of directories in the working directory.
         """
         # Image filelist has changed, relevant for thumbnail and image mode
+        images, directories = files.strip_meta((images_meta, directories_meta))
```
karlch, line 209: This links to the comment made in the library, we end up with calls that expect the meta but don't get it.

vimiv/gui/library.py
```diff
@@ -468,17 +478,17 @@ def _add_rows(self, paths: List[str], are_directories: bool = False):
...
         for i, path in enumerate(paths, start=self.rowCount() + 1):
-            name = os.path.basename(path)
+            name = os.path.basename(path[0])
             if are_directories:
                 name = utils.add_html(name + "/", "b")
             if path in api.mark.paths:
                 name = mark_prefix + name
             with contextlib.suppress(FileNotFoundError):  # Deleted in the meantime
-                size = get_size(path)
+                size = get_size(*path)
```
karlch, line 487: This seems to break when we load a changed directory as the working directory handler strips the meta and we end up with a simple list of strings, not a list of tuples.

karlch (commented):
Thanks again for all the work that went into this. Currently there seems to be a bit of a mix with calls that expect the meta, and calls that don't, leading to quite some bugs.

Personally I am not sure if I would include this PR at all. The added complexity of passing tuples instead of simple paths is quite large and does make the already quite involved loading process even more confusing. Besides this, using the low level `stat` calls is confusing to me and a lot less readable, but this is not really a strong argument as this could be learned / adapted. Overall I am just not sure if the benefit is large enough to justify the added difficulties.

woefe:
I fully agree with your judgement. The loading process is quite complex and this PR just adds more complicated stuff on top. I guess refactoring or properly fixing this PR will take quite some effort. And in the end I still waited 25 seconds for vimiv to start. Don't feel sorry if you close it without merging ;) I won't find the time to work on this. Mainly, I just wanted to see if there are some easy improvements, but I guess that's not the case.

Time is probably spent better elsewhere. I saw the idea of lazy-loading the library, for example. For sure the performance impacts of lazy-loading will be far greater than this PR (although I'm certain that it will cause large amounts of new issues as well, since the library loading is so intertwined with the core of vimiv).

karlch:
Thanks for your judgment, I will then close this PR, maybe it can serve as a useful reference for future attempts :blush: 

Yes, i fully agree, the lazy-loading is a lot more involved than it may seem due to the intertwining you mention. So all-in-all everything would probably be quite a major effort, which I currently also don't have the time for :disappointed:

PR #454, opened 2021-12-26
Add missing metavar for --output argument

jcjgraf:
missed that in #381

P.S. I am not sure where the `optional argument` -> `options` change comes from... Probably something was changed but the man page not rebuild, I guess

### DIFF
--- a/misc/vimiv.1
+++ b/misc/vimiv.1
@@ -1,8 +1,5 @@
 .\" Man page generated from reStructuredText.
 .
-.TH "VIMIV" "1" "May 05, 2021" "" "vimiv"
-.SH NAME
-vimiv \- an image viewer with vim-like keybindings
 .
 .nr rst2man-indent-level 0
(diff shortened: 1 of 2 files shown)


karlch:
Thanks for catching this! I assume the optional arguments -> options change is due to [this change in python 3.10](https://bugs.python.org/issue9694). I guess we should also stick to the "official" wording, although our man sectioning is now a bit weird with `OPTIONS` as the parent and `options` as one child. Maybe we can think of a better name for the parent section `OPTIONS`?

jcjgraf:
Something we could possibly do is to get rid of the *OPTIONS* parent title altogether and use the two children at the top level.

But anyways, I think we should either do that in #390 or after it has been merged. Therefore, I have undone this optional arguments -> options change.

karlch:
Thanks, I agree this would best be done in another separate commit / PR.

PR #467, opened 2022-02-25
Move Metadata Support to Pluginspace

jcjgraf:
Desperate attempt to get metadata to work nr 148 :yum: 

We have had so many troubles to get metadata support and none of the available solutions seemed perfect. To quickly summarize to issues we have had:
- No formatted output
- Limited image type support
- Not in official repository
- Not even in aur
- Many build deps
- Generally broken
- Deps break CI
- Package no longer maintained
- etc.

So I thought that is is probably the best idea to move the "dirty code" out to plugin space. More precisely, my idea was to remove all metadata related code which is fixed to a certain package, and instead add an interface which allows to load plugins which add metadata support. I would create ~one~two "official"/internal plugin for metadata support based on piexif **and pyexiv2**, and then anyone is free to implement metadata plugins using whatever method one likes.

~This PR is very WIP.~ *Almost complete*. Any feedback is very welcome

**Update:** I will maintain the TODOs in order to keep the overview...

## TODOs

- [x] Main Implementation
    - [x] Abstract class with interfaces for for `get_metadata`, `get_keys`, `copy_metadata`, and `get_date_time`
    - [x] Metadata plugin registration functionality

- [x] Metadata Plugins
    - [x] With `piexif` backend
        - [x] Provide backward compatibility for having the long keys
    - [x] With `pyexiv2` backend
        - [x] Do not provide backward compatibility for having short keys, as this is really ugly and confusing

- [x] Configuration
    - [x] Default backed loading
        - For backward compatibility, a default backed should be loaded
        - [x]  Handled by plugin `metadata`
            - [x] Should be possible overwrite default and have (only) a specific plugin loaded
                - The default backend should not be loaded at all
            - [x] Should be possible overwrite default and have not metadata plugin loaded
    - [x] Should keys be case-insensitive?
        - Currently they are not, and I do not see a reason why they should not be (but maybe there is?)

- [x] Handle case where there is no metadata support
    - [x] `MetadataWidget` should not be loaded
        - Done using the `plugins_loaded` signal (thanks @karlch!)

- [ ] Testing
    - [x] Have been updated (thanks @karlch!)
    - [ ] More Tests (@karlch)?
    - [ ] Anything else (@karlch)?
    - [ ] There are still a few suboptimal things. See:
> The integration tests unfortunately still require both marking and using the corresponding fixture, in case we test piexif / pyexiv2 explicitly as I couldn't figure out how to mark a fixture (instead of the tests) quickly. Could definitely be improved.

- [x] CI
     - [x] pyexiv2 has been removed (in a seperate PR)
     - [ ] Anything else (@karlch)?

- [x] Docs
    - [x] Detailed section about metadata
    - [x] Change *exif* to *metadata* where applicable
    - [x] How should user metadata plugins be listed?
    - [x] Document default behaviour (what happens when no backend is explicitely loaded)

- [x] Misc
    - [x] Change *exif* to *metadata* in code and comments where applicable

## Deferred for another PR
- Internal metadata keys (ref #325)
    - Will be implemented as a separate metadata plugin
    - No special functionality (interface) needed to make it work

- Display of metadata plugin (backend) version in `:version/--version`
    - [x] Interface (`imutils.metadata.get_registrations`) has been added

- Deprecate default metadata support?

- Interface for writing metadata


karlch:
While I won't be able to do a detailed code review quite yet, I do have some initial thoughts (that we probably have shared over mail at some point) and at least CI / dependencies have been fixed in master by now, so a quick rebase would be great :blush:

I absolutely love the idea of getting the metadata havoc out of mainline, it has been quite an ordeal, you are right, and thanks again for all your hard work on this! However, I do believe we somehow need to support pyexiv2, given we asked users to switch "recently". Maybe you could just quickly recycle the existing code into another plugin with corresponding name, so we have both in the codebase for now? With this + the rebase, I should absolutely do a full review and we can discuss the details.

Later on I can take care of any CI / testing if you like, I feel like this will be messy :joy: but currently all should be fine, as we never test against pyexiv2 (on CI).

jcjgraf:
I have rebased for latest Master. Also, I have rewritten most code, as my initial implementation was a bit limiting, as it allowed only for one metadata handler to be present at the time.

The new implementation is a bit more complex in therms of code, but gives more freedom to users, in regards on how they want to handle metadata. 
Using `@exif.register(exif.Methods.XXX)` users can register a function as a certain implementations of a metadata function (i.e for `get_formatted_metadata`, `get_keys`, `get_date_time`, `copy_metadata`). For all functions, with the exception of `get_date_time`, multiple implementations can be registered. The output of the gets combined.
This approach trivially solves #325, as I can create simply create another plugin, that registers some functions for `get_keys` and `get_formatted_metadata`.

`exif.MetadataHandler` is the class used to access metadata of an image (e.g. by the MetadataWidget). When one of its methods is called, `exif._MetadataRegistry` is accessed to load all available function implementations (`@exif.register` stores implementations in this registry).

As suggested, I have also created a plugin for metadata support using py3exif2 (which is, however, completely untested (and might even crash) as I am not able to get py3exif2 to run currently).

The latest commit is very VIP. I still need to do a lot of cleanup, fix docstrings, fix namings, and minor todos. Also, I have not looked into testing/CI yet (therefore the CI also fails).
However, it would be great if you could provide me with some (high-level) feedback. How you like the general idea and code pattern, or can you come up with something more slick?

karlch:
This is golden! I'll only leave a quick overall impression, not any detailed inline comments as of now then.

First of all, I absolutely love the flexibility this gives. So following are some random notes, in no particular order:
- There are other places around the code (`imutils`) that use `ExifHandler`, these should also use `MetadataHandler` I guess.
- All the `if len(...) == 0` checks could be replaced by a simple `if ...`, but even then I guess we will need a wrapper for clients, i.e. `self.has_formatted_metadata` as property of the registry class or even as module-level functions, which would allow `gui.metadatawidget` call `exif.has_formatted_metadata()` in advance (assuming the importing is done in the correct order).
- In general, I like how flexible the registry decorators are, but do we need them all? I guess at least the raw metadata part can be removed, as it is unused.
- I guess the metadata retrieval of the current path should be cached in each plugin. Probably a simple `functools.lru_cache(1)` is enough.

Random thoughts over :D Once again, love it, and do think the overall structure makes a lot of sense. (Code) details to be checked later :blush:

jcjgraf:
Thanks for the high-level feedback. I am glad you like the overall structure.

Some comments on your points:

1. True, thanks!
2.
    a. Often I prefer to be explicit and avert using truthy/falsy values as it can lead to nasty bugs. But I will change it.
    b. I like the idea of using module-level functions. I tried something similar, but the issue is that `gui.metadatawidget` is loaded before the plugins are loaded (`startup.py` imports `gui.mainwindows`, which imports the `gui.metadatawidget`, while plugins are loaded in `setup_pre_app`). Do you know of a good way to fix that?
3. True, `get_raw_metadata` is not used yet. My idea was that this method should provide an interface for plugins that need access to (raw) metadata. Also, piexif does not provided formatting for metadata, so it should theoretically implement `get_raw_metadata`. However, this raises the question on how to decide what to display in the MetadataWidget. I have also though about renaming the methods to `get_metadata_for_widget` and `get_metadata_internal` (or something the like). What do you think?
4. Oh, this sounds great. I was not aware of this decorator. Thanks for the hint!

Some additional thoughts:
- I think the configuration of the sets still makes sense. Or do you see anything we should change about that?
- Concerning testing, what should be tested?

> Later on I can take care of any CI / testing if you like, [...]
I will have a look at CI/testing, but I cannot promise anything. Maybe it will be better if you sort that out 😬

I will work on it in the following days. Thanks again for your feedback. I wish you a nice weekend!

jcjgraf:
Thanks for the high-level feedback. I am glad you like the overall structure.

Some comments on your points:

1. True, thanks!
2. 
    a) Often I prefer to be explicit and avert using truthy/falsy values as it can lead to nasty bugs. But I will change it.
    b) I like the idea of using module-level functions. I tried something similar, but the issue is that `gui.metadatawidget` is loaded before the plugins are loaded (`startup.py` imports `gui.mainwindows`, which imports the `gui.metadatawidget`, while plugins are loaded in `setup_pre_app`). Do you know of a good way to fix that?
3. True, `get_raw_metadata` is not used yet. My idea was that this method should provide an interface for plugins that need access to (raw) metadata. Also, piexif does not provided formatting for metadata, so it should theoretically implement `get_raw_metadata`. However, this raises the question on how to decide what to display in the MetadataWidget. I have also though about renaming the methods to `get_metadata_for_widget` and `get_metadata_internal` (or something the like). What do you think?
4. Oh, this sounds great. I was not aware of this decorator. Thanks for the hint!

Some additional thoughts:
- I think the configuration of the sets still makes sense. Or do you see anything we should change about that?
- Concerning testing, what should be tested?

> Later on I can take care of any CI / testing if you like, [...]
I will have a look at CI/testing, but I cannot promise anything. Maybe it will be better if you sort that out 😬

I will work on it in the following days. Thanks again for your feedback. I wish you a nice weekend!

karlch:
2. a) Understandable, I just feel like the implicit version is so much more readable in python.
b) Unfortunately that is what I feared. One of the rare occasions I miss compiling vs dynamic importing :joy: Feel free to just write the `has_...` function(s) and leave the `while True` in `gui.metadatawidget` for now. I can play around with the import order in the same step as the CI / testing havoc, sounds like fun :blush: Difficult to say what needs to be done, without just doing it by changing the lines of code ..
3. I see where you are coming from now. Do the various plugins provide the "raw" metadata in consistent format? If not, it would be really hard for another plugin to link in at that point, without assuming the backend. In which case it might as well just use the backend directly ... I also prefer the current naming, the new naming would mean that the exif plugins already know what they are used for, seems a bit backwards to me :blush:

Concerning your additional thoughts:
- If this refers to the whole `[METADATA]` part in the config section, I agree, it still makes a lot of sense and is very powerful as-is.
- Ideally at least everything that is currently tested :smile: but you can definitely leave that part out for now if you like, and I can sort it out later, as I can't quite grasp how much havoc it will be just yet.

Thanks again for tackling this! And a great week-end to you too! :partying_face:

jcjgraf:
Thanks for you respond. Concerning point 3.:

> Do the various plugins provide the "raw" metadata in consistent format?
The way metadata are stored in images is (obviously) standardized (by Exif, IPTC and XMP), however, each package can present the data in whatever way they like. But most fields are basic types anyways. In the worst case a simple regex is probably sufficient to extract the data from the string value, provided by the backend.

> [...] it might as well just use the backend directly
If somehow possible, I would like to prevent that any (future) plugin that requires metadata support, has to deal with the backend (piexif, exiv2, py3exiv2, etc.) themselves directly. But I vision that `imutils.Metadata` provides and interface/API for these plugins to get all metadata functionality they need. And this functionality should include:
- reading formatted data (for direct display)
- reading raw data (for further processing)
- and the more I think about it, probably also writing data to fields (e.g. for a plugin that allows for setting star ratings).

I hope that clarifies my intend a little. Let me know if this makes sense to you and what you think about it.

karlch:
That definitely makes sense, I hope the work that goes into this doesn't get out of hand though :joy:

jcjgraf:
Code-vise, this PR is pretty much done;

I decided to get rid of the `get_raw_metadata` method and I also did not implement any functionality for writing metadata. I think it is better to deal with that once there is actual demand for it. Since currently, I am not sure that the exact requirements are.

Concerning the MetadataWidget, I do not think that simply by reordering imports we are able to fix the issue. I guess some kind of callback function is needed to set the MetadataWidget as an overlay after the plugins are loaded. But I guess you will come up with a better solution than me, therefore, I will leave this up to you :blush:

I am also not sure what information concerning metadata support to return in `version.info`.

All tests are currently broken. I will look into them and try to fix them. Did I get it correctly, we get rid of all py3exiv2 related tests? So we only test the no-metadata case and when metadata_piexif is loaded?
But as said, probably you need to help me out with that :grimacing:  I do also not know that needs to be changed with the CI. Currently I get some linter and mypy errors that pyexiv2 cannot be imported.

I will update the docs and the changelog maybe later today or in the next few days.
Do you already know when you have time for a detailed review? You could also already review the code now if you like.

karlch:
Thanks again! :blush: 

This makes perfect sense, the whole PR is complex enough as-is.

You are probably right, in some way the metadata widget has to be imported after the plugins, which may mean it is added to the overlays after mw is already created.

That is a good question :thinking: Could we add a "name" and "version number" to the metadata plugins somehow, and this is then displayed?

Concerning the tests: yes, currently pyexiv2 is not tested, at least not on the CI. I can still run it locally though, since for some reason it still runs fine on my laptop :smile: With the imports, it probably still makes sense to lazy-import piexif / pyexiv2? Not sure about the errors though, will look into that later.

I'll try to keep the current momentum up, hopefully at latest next week-end in the train :blush:

jcjgraf:
> Could we add a "name" and "version number" to the metadata plugins somehow

Do you mean name/version of the plugin or of the backend? I have thought about adding the name of the plugin too, but I was not able to come up with a clean solution that enforces clients to set this value. Using some `inspect.stack()` hackery, one can get the plugin name (by extracting it from the path of the executed file). Though, that is probably not very reliable.
I have thought a lot about having a Handler class that plugins can subtype (instead of having the register decorators). Maybe that is the way to go? An abstract class is not really applicable, as clients are not required to implement all methods.

> With the imports, it probably still makes sense to lazy-import piexif / pyexiv2?

You mean in the two metadata plugins? Sure, I can do that.

Sure, no pressure at all! I find it great that you are able to invest again some time for vimiv! :smiley:

karlch:
I also had some thoughts on the class that clients can subtype. We would then somehow have to figure out what they implemented, some funky inspection I guess, and add to the implementation registry as is currently the case upon construction of the parent class. I.e. something like (does not work, don't take the names too seriously, just brainstorming):
```python
class MetadataPiexif(MetadataBase):
     def __init__(self):
          super().__init__("piexif", piexif.version)
          
class MetadataBase:
    def __init__(self, name: str, version: str):
        self.name = name
        self.version = version
        # Check what the child implements and update registry
        
    def get_formatted_exif(self):
        raise NotImplementedError("Must be implemented by the child class")
```
Could work :thinking: 

Thanks! Only if this avoids the actual import though :laughing: i.e. there is no actual call to piexif.* upon startup. This would probably break at latest if we call `piexif.version` or so :cry:

jcjgraf:
Yeah, something like this should work. But as you say yourself, we would need to detect what methods get implemented (/overwritten).
Alternatively, we could redesign the registry, to not store list of function for each operation, but the subtypes of the plugin directly. For each of its methods, the `MetadataHandler` would then iterate over all subtypes, and executed their respective method. If the client did not implement that operation, the `MetadataHandler` gets a `NotImplementedError` which it can ignore.

karlch:
Makes sense, looking at these two options, the baseclass - subtyping approach seems the most clean. Especially with your option of ignoring the `NotImplementedError` :blush:

jcjgraf:
Doing this, however raises the question on how to implement `has_get_metadata()`. It seems like we need to inspect the subclass anyways to see what methods they implement :thinking:

karlch:
Yep, you are of course right, this will need inspecting anyway ...

jcjgraf:
Could you by any chance take a look at the latest commit? I have refactored the registration process such that clients are required to subclass `MetadataPlugin`, as discussed.
The name and version are not specified in the init function, as you suggested, but as class attributes. This has the advantage that I can access these values in `metadata.register`, before creating any instance of the subtype class. Using `MetadataPlugin.__init_subclass__` I enforce that clients set these two values.

The commit is very WIP. I have only adopted the piexif plugin so far.

From a high-level, this this look good to you?

vimiv/imutils/metadata.py
```diff
@@ -0,0 +1,304 @@
...
+        for _, backend in _registry.items():
+            with contextlib.suppress(NotImplementedError):
+                out = itertools.chain(backend(self._path).get_keys(), out)
+
+        return out
+
+    @staticmethod
+    def raise_exception() -> NoReturn:
+        """Raise an exception for a operations without implementation."""
+        msg = f"{inspect.stack()[1][3]} has no implementation"
```
karlch, line 262: Random note flying over this, I would really avoid using the stack here :laughing: Maybe just call `raise_exception` with a string? Or, if you really want to avoid re-writing the name, decorate these methods with a decorator somewhere along:
```python
def metadata_function(check):
    def decorator(func):
        def inner(self, *args, **kwargs):
            if not check():
                self.raise_exception(func.__name__)
            return func(self, *args, **kwargs)
        return inner
    return decorator
```
but even this is probably over the top.

karlch:
I like this a lot! The structure seems very clean, definitely nicer than everything we had so far, thanks!! :blush: 

Sorry for all the random inline comments, I made one .. and then another .. and then it got out of hand :smile: 

Would moving `name` and `version` into a property, i.e.
```
@property
def name(self):
     return "piexif"
```
as mentioned briefly in an in-line comment also work? This would avoid the direct call to `*.version` and thus the `import`, in case we go back to lazy loading, I hope. For user-plugins, that is up to them if it is relevant, but ideally ours are as fast as possible :blush:

vimiv/imutils/metadata.py
```diff
@@ -0,0 +1,304 @@
...
+
+
+def register(plugin: Type[MetadataPlugin]) -> None:
+    """Register metadata functionality for usage by MetadataHandler.
+
+    Args:
+        plugin: Subclass of MetadataPlugin that implements metadat functionality.
+    """
+    assert plugin.name, "Required to have `plugin.name` set"
+    assert plugin.version, "Required to have `plugin.version` set"
```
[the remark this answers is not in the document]
jcjgraf: They are not needed. They are just for the static analysis. I can also just ignore the warning if you prefere:
```
_registry[plugin.name] = plugin  # type: ignore[index]
```

vimiv/plugins/metadata_pyexiv2.py
```diff
@@ -0,0 +1,126 @@
...
+
+    def get_keys(self) -> Iterable[str]:
+        """Retrieve the key of all metadata values available in the current image."""
+        if self._metadata is None:
+            return iter([])
+
+        return (key for key in self._metadata if not is_hex(key.rpartition(".")[2]))
+
+
+def init(*_args: Any, **_kwargs: Any) -> None:
```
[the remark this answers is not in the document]
jcjgraf: Thanks for catching. It should actually not be empty, I just forgot to register the plugin (did not test it as I am not able to get pyexiv2 to work...) :grimacing:

vimiv/imutils/metadata.py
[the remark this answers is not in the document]
karlch: Ah, because `mypy` doesn't know `MetadataPlugin` must actually defines these I guess? Would probably `type: ignore` then with a quick comment explaining this :blush:

vimiv/plugins/metadata_pyexiv2.py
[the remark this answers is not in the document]
karlch: Random question, I guess building vimiv with tox also doesn't work with pyexiv2 for you? I.e. adding both the piexif and the pyexiv2 requirements file to the mkvenv tox environment, running `tox -e mkvenv` and then vimiv from `.venv/bin/vimiv`?

jcjgraf:
Thanks for the promptly feedback! I am glad you like the structure :blush:

I will look into the points you have mentioned. Especially into the lazy-loading and converting the attributes into properties.
This actually makes me reconsider the use of an abstract class, as:
- `abc` can enforce the implementation of properties (but not of attributes, therefore my hack using `__init_subclass`).
- I have the feeling that we give user a little too much flexibility :thinking:  I think every plugin should be required to implement `MetadataPlugin.get_metadata` and `MetadataPlugin.get_keys`. Or can you come up with a legitimate case where this is not applicable? This also simplifies the metadata support check in `gui.metadatawidget`, as with the current code, we could have a registered function for get_metadat, but not for get_keys (which is used in `metadata_list_keys`).

I have also already looked **a little** into fixing the tests. A thing I am not really sure about is, that for testing I need to force the `_registry` into a certain state (e.g. ensure that only the piexif plugin is loaded). This is something what is not required during normal operation. Hence, should I just access the private `_registry` (which leads to a linter error), or create some extra functionality in `imutils.metadata` that is only used for testing?

karlch:
Good point, with abstract properties this would actually be pretty clean :blush: 

I was thinking about this as well, the internal metadata stuff would essentially implement only these two, but definitely these, right? In this case no, I cannot think of a use-case, and these could then also be abstract. Although internal keys may not need the filename, but could ignore it just for the sake of proper overriding.

Requiring a certain state for the tests makes perfect sense. Random comment: make sure to do this in a proper fixture, that resets the registry to the previous state after the test. You can e.g. refer to the `test_read_bindings.py` integration test which has
```python
@pytest.fixture(autouse=True)
def reset_to_default(cleanup_helper):
    """Fixture to ensure everything is reset to default after testing."""
    yield from cleanup_helper(api.keybindings._registry)
    api.settings.reset()
```
Accessing private attribute in tests should be perfectly fine, `protected-access` is disabled for tests in `scripts/lint_tests.py` :blush:

jcjgraf:
>I was thinking about this as well, the internal metadata stuff would essentially implement only these two, but definitely these, right?

Exactly, I have already already created a internal metadata plugin by recycling the code from #325 . I have not pushed it, as I though I would create a seperate PR for that. Or should I include in  in here?

Thanks for your hints concerning testing :smiley:

karlch:
I would also keep the internal part in a separate PR, was just the only other plugin that came to my mind :laughing:

vimiv/imutils/metadata.py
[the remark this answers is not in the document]
jcjgraf: Yes exactly.

Is there any particular reason why you do not specify the exact code/error to ignore? I.e. why `type: ignore` over `type: ignore[index]`?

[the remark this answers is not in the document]
karlch: Bad habits if anything in the code, laziness when typing here :laughing:

vimiv/imutils/metadata.py
jcjgraf: fixed in aafe415fb7230ed7b63342eee41de7eeafdb190d

vimiv/imutils/metadata.py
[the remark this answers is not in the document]
jcjgraf: Haha, makes sense :upside_down_face: 

This is no longer relevant due to the refactoring in aafe415fb7230ed7b63342eee41de7eeafdb190d

vimiv/plugins/metadata_piexif.py
```diff
@@ -0,0 +1,139 @@
...
+from vimiv.utils import log
+
+_logger = log.module_logger(__name__)
+
+
+class MetadataPiexif(metadata.MetadataPlugin):
+    """Provided metadata support based on piexif."""
+
+    name = "piexif"
+    version = piexif.VERSION
```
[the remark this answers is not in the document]
jcjgraf: Should be considered. Could you please verify I did everything correctly with the lazy loaying? :blush:

jcjgraf:
I have refactored the `MetadataPlugin` to make it an abstract class. This has definitely improved the structure again. I think the implementation is now as clean as possible :laughing:
See aafe415fb7230ed7b63342eee41de7eeafdb190d and its commit message for more details on what has changed.

The only not-so-nice thing I see is, since `name` and `version` are now properties (and there is nothing like an abstractclassproperty), I need to create an instance of each metadata plugin implementation, to extract these values (e.g. for generating the version output). See `imutils.metadata.get_registrations`. Do you see any better option for this?

Could you please also verify that I did everything correctly with the lazy loading of the backends? I am not sure how to practically verify it :blush:

Code-vise, I am pretty much done. One of the few things left is, to figure out what to do about the MetadataWidget. :thinking:

I will look into testing and updating the documentation in the next few days.

Btw, can you tell me why locally running `tox -e mypy` reports issues, that are not reported by the CI? :thinking:

vimiv/plugins/metadata_pyexiv2.py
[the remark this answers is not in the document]
jcjgraf: Thanks for the hint, this seems to be working actually. How weird :confused:

I have not tried it for quite some time, maybe something has changed. I used to get an error related to some shared library when trying to import pyexiv2 after installing it.

jcjgraf:
I have made a first attempt on fixing the tests (well, I only looked at conftest and test_version so far). It would be great if you could give me some feedback. In particular:

- The `metadata_markers` (former_`exif_markers`) have been adapted. However, I do not think that the conditions make sense. For `metadata` and `nometadata` the condition only indicates whether the current instance of vimiv has metadata support (or not). While for `piexif` and `pyexiv2`they indicate if the backend package is installed. What exactly should the condition be?

- In addition, having condition `metadata_pyexiv2.pyexiv2 is not None` fails for mark `pyexiv2`. I am not sure why that is the case.

- Fixtures `metadata_support`, `no_metadata_support`, `piexif` and `pyexiv2` have been added and implemented. They bring `metadata._registry` into the respective state, and make use of `reset_metadata_registration`, to undo the changes after each tests.

- Updated `test_version`. Am I required to set for each testcase the mark, as well as appropriated fixture (i.e. `@mark.piexif` and specify `test_XXX(piexif)`. This seems a bit tedious.

Sorry for all these questions :grimacing: I just want to make sure everything is sorted out, before digging deeper into the testcases.

karlch:
Nice work! I will probably not be able to look into this anymore before the week-end though :blush:

jcjgraf:
No worries. Thanks for letting me know!

Have a nice and successful week

vimiv/plugins/metadata_piexif.py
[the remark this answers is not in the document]
karlch: Looks good to me, thanks :+1:

vimiv/version.py
```diff
@@ -42,8 +41,9 @@ def info() -> str:
         f"Qt: {QT_VERSION_STR}\n"
         f"PyQt: {PYQT_VERSION_STR}\n\n"
         f"Svg Support: {bool(QtSvg)}\n"
-        f"Pyexiv2: {exif.pyexiv2.__version__ if exif.pyexiv2 is not None else None}\n"
-        f"Piexif: {exif.piexif.VERSION if exif.piexif is not None else None}"
+        # TODO: what to do about them?
+        # f"Pyexiv2: {exif.pyexiv2.__version__ if exif.pyexiv2 is not None else None}\n"
+        # f"Piexif: {exif.piexif.VERSION if exif.piexif is not None else None}"
```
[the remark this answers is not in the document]
karlch: Yes, but unfortunately `--version` does not know there is metadata support, as plugins are loaded later ... I am currently not sure how to solve this, but would probably go hand-in-hand with the current issue in `gui.metadata`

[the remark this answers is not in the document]
jcjgraf: I see. I did not think about `--version`, only about `:version`

karlch:
I'll jump right in with some higher-level comments, before I dig into the code and leave some more inline comments :blush: 

You are right, although not as nice as properties, I do agree having abstract classmethods makes sense here, to avoid creating instances which are not needed :+1: 

Lazy importing looks good to me. Random note for future: I run `python -X importtime -m vimiv` and check if e.g. `piexif` ends up in the import log, which it shouldn't until we do something metadata related :blush:

I'll go through the test-related stuff one-by-one and leave inline comments where applicable.

vimiv/gui/metadatawidget.py
```diff
@@ -13,27 +13,29 @@
...
+from vimiv.imutils import metadata
 from vimiv.config import styles
 
 _logger = utils.log.module_logger(__name__)
 
 
-if exif.has_exif_support:
+# TODO: conditional fails as this file is imported before plugins are loaded
+# if metadata.has_get_metadata():
+if True:
```
karlch, line 24: Adding comment as a reminder that this is still on the todo list :blush:

vimiv/imutils/metadata.py
```diff
@@ -0,0 +1,284 @@
...
+`MetadataHandler` is used to interact with the metadata of the current image. It
+relies on the functionality provided by optionally loaded metadata plugins. Such
+plugins implements the `MetadataPlugin` abstract class and registers that class
+using the `register` function.
+
+Module Attributes:
+    _registry: List of registered `MetadataPlugin` implementations.
+"""
+
+from abc import ABC, abstractmethod
```
karlch, line 19: Please use `import abc` for consistency :blush:

vimiv/imutils/metadata.py
```diff
@@ -0,0 +1,284 @@
...
+    @abstractmethod
+    def name(self) -> str:
+        """Get the name of the used backend.
+
+        If no backend is used, return the name of the plugin.
+        """
+
+    @property
+    @abstractmethod
+    def version(self) -> str:
```
karlch, line 59: Fully agree with making these abstract classemethods or staticmethods, leaving a comment so we have a place to tick it off :blush:

vimiv/imutils/metadata.py
```diff
@@ -0,0 +1,284 @@
...
+        """
+        raise NotImplementedError
+
+    def get_date_time(self) -> str:
+        """Get creation date and time of the current image as formatted string."""
+        raise NotImplementedError
+
+
+# Stores all registered metadata implementations.
+_registry: List[Type[MetadataPlugin]] = []
```
karlch, line 100: Type annotation should be `List[MetadataPlugin]`, no?

vimiv/imutils/metadata.py
```diff
@@ -0,0 +1,284 @@
...
+    @property
+    def has_copy_metadata(self) -> bool:
+        """True iff `MetadataHandler` has an implementation for `copy_metadata`."""
+        return any(
+            list(
+                map(
+                    lambda e: e.copy_metadata != MetadataPlugin.copy_metadata, _registry
+                )
+            )
+        )
```
karlch, line 131: This looks funky :blush: is it not the same as
```python
any(e.copy_metadata != Metadata.copy_metadata for e in _registry)
```
?

vimiv/imutils/metadata.py
```diff
@@ -0,0 +1,284 @@
...
+    @property
+    def has_get_date_time(self) -> bool:
+        """True iff `MetadataHandler` has an implementation for `get_date_time`."""
+        return any(
+            list(
+                map(
+                    lambda e: e.get_date_time != MetadataPlugin.get_date_time, _registry
+                )
+            )
+        )
```
karlch, line 142: Same as previous, much easier with a generator expression in the any call.

vimiv/imutils/metadata.py
```diff
@@ -0,0 +1,284 @@
...
+            Dictionary with retrieved metadata.
+        """
+        out: MetadataDictT = {}
+
+        for backend in _registry:
+            with contextlib.suppress(NotImplementedError):
+                # TODO: from 3.9 on use: c = a | b
+                out = {**backend(self._path).get_metadata(keys), **out}
+
+        # TODO: ensure keys are returned in the right order
```
karlch, line 163: What would the "right" order be? Ideally they don't overlap, but if they do, just the user-defined order, no?

vimiv/plugins/metadata_piexif.py
```diff
@@ -0,0 +1,149 @@
...
+    def version(self) -> str:
+        """Get the version of the used backend."""
+        return piexif.VERSION
+
+    def get_metadata(self, desired_keys: Sequence[str]) -> metadata.MetadataDictT:
+        """Get value of all desired keys for the current image."""
+        out = {}
+
+        # TODO: remove
+        desired_keys = [key.rpartition(".")[2] for key in desired_keys]
```
karlch, line 66: Remove as this is filtered later on?

vimiv/plugins/metadata_pyexiv2.py
```diff
@@ -0,0 +1,134 @@
...
+        """Get value of all desired keys for the current image."""
+
+        out = {}
+
+        if self._metadata is None:
+            return {}
+
+        for base_key in desired_keys:
+            # TODO: potentially remove
+            # For backwards compability, assume it has one of the following prefixes
```
karlch, line 60: This goes hand in hand with the previous. I suppose it makes sense to get everything here, and finally filter in the metadata widget?

karlch, line 60: This goes hand in hand with the previous. I suppose it makes sense to get everything here, and finally filter in the metadata widget?

vimiv/version.py
```diff
@@ -36,14 +36,19 @@ def info() -> str:
         if git_info is not None
         else f"vimiv v{vimiv.__version__}"
     )
+    metadata_plugins = metadata.get_registrations()
+    metadata_plugins_version = (
+        "False"
+        if not metadata_plugins
+        else "".join(f"\n    {n}: {v if v else '-'}" for n, v in metadata_plugins)
```
karlch, line 43: When would version ever be falsy? I would just assume that this is defined as a string

pytest.ini
```diff
@@ -5,9 +5,9 @@ faulthandler_timeout = 30
 markers =
     current: Mark tests during development
     imageformats: Require retrieving images from the web to test additional formats
-    exif: Require exif support
+    metadata: Require metadata support
```
karlch, line 8: I will need more time with the tests, as I am currently unsure how to best set the markers and what to test where. As there is some train chaos (as usual :joy:) I will probably not make it today, and might end up pushing some commits instead of leaving a lot of comments for simplicity which you can cherry-pick, if you don't mind?

Note to self:
- update marker in tests, also end2end feature files
- think about `metadata` and `nometadata` markers, and definitely fixtures, it *should* be enough to mark the corresponding tests correctly, instead of forcing the code into a specific state via a fixture :thinking: This would also solve the version part, only test for a specific string if it is marked accordingly. Alternatively mocking could make sense as well. Processing :blush:

karlch (changes requested):
The code is indeed in a nice and clean state, the tests and the import are the two blocking issues I will look into directly next, but leaving a few (simple) comments here before my train internet crashes :blush: happy hacking and have a great week-end!

vimiv/imutils/metadata.py
jcjgraf: No, because plugins call `metadata.register` by passing the implementation of `MetadadaPlugin` as the argument (i.e. the class itself). And while the type of instances of the class `MetadataPlugin` is `MetadadaPlugin`, the class itself is of type `Type[MetadadaPlugin]`.

karlch: Sorry for the noise then :blush:

vimiv/version.py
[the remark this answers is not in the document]
karlch: Honestly, I think we can just remove this part from `--version`, having plugins this early is just complete havoc. And having just this one here seems inconsistent, we also don't indicate print support for example.

vimiv/gui/metadatawidget.py
karlch: I pushed an idea to `core/metadataplugin-jcjgraf`, what do you think about solving this with a `plugins_loaded` signal? I have the feeling this could be useful in general :thinking: 

But definitely doesn't help with `--version` ...

karlch: I pushed an idea to `core/metadataplugin-jcjgraf`, what do you think about solving this with a `plugins_loaded` signal? I have the feeling this could be useful in general :thinking: 

But definitely doesn't help with `--version` ...

vimiv/plugins/metadata_piexif.py
jcjgraf: Sorry for not providing a proper ToDo message. This conversation, and the subsequent are directly related to each other. Though, they have nothing to do with any sort of filtering.

Let me elaborate a bit...
pyexiv2 and piexif use different keys to index the same data. The former one used a `Group.Subgroup.Keyname` key, while the later one uses simply `Keyname` (actually, it also uses some kind of prefix, but it is off the standard).
When adding pyexiv2, we thought about backwards compatibility in the following sense:
- Backwards Compatibility Nr 1: Someone having the old piexif keys defined in their config, but wanting to use the (at that time) new pyexiv2 backend.
    - Solution: We *predict* the `Group.Subgroup` prefix to be one of the very common ones.
- Backwards Compatibility Nr 2: Someone having the new pyexiv2 keys in their config (which are the default), but wanting to use the old piexif.
    - Solution: Strip `Group.Subgroup` prefix

These two ToDos are referring to this backwards compatibility. The issue with maintaining the backwards compatibility is, in my view, that it may be a bit confusing; The keys defined in the config do not match with the ones returned by `:metadata-key-list`. Nevertheless, the widget is able to retrieve and display some data for this key. Also, the behaviour is (so far) not documented.

Personally, I do not think that is makes sense to maintain backwards compatibility Nr 1 any longer.However, for Nr 2, it probably makes sense to keep it as is. This way one does not need to modify the metadata keys in the config in order to have metadata support using the piexif based plugin.

What do you reckon?

vimiv/version.py
jcjgraf: When the plugin does not require any backend. Thich is for example the case for the internal metadata plugin. But maybe we should just ignore this special cases :thinking:

vimiv/plugins/metadata_piexif.py
karlch: Ah, thanks for your clarification, I somehow thought you wanted to remove the `desired_keys` part in general.

This makes perfect sense, I agree with removing Nr. 1 and keeping Nr. 2 :+1:

vimiv/version.py
karlch: It will still have to override the two abstract methods, no? So this one could e.g. return "internal" as name, and "N/A" as version

vimiv/plugins/metadata_pyexiv2.py
jcjgraf: duplicate. Probably the train interent screwed you? :laughing:

karlch: Funky, I definitely don't remember writing this twice, so I shall blame the train :laughing:

karlch: Funky, I definitely don't remember writing this twice, so I shall blame the train :laughing:

pytest.ini
jcjgraf: This makes sense. Feel free to push to this branch.

I am glad you can help with sorting out the setup. Once this is done I will continue with fixing, and potentially adding, tests. :blush:

And I wish you a good journey! :steam_locomotive:

vimiv/imutils/metadata.py
jcjgraf: Haha true! Sometimes I get carried away with my thought about functional programming :upside_down_face:

pytest.ini
karlch: So I pushed some stuff :blush: not super happy with the state, but at least there is some progress. I believe having the metadata test as integration instead of unit test makes sense, as we then don't have to go through the havoc of mocking, but can test the integration of `imutils.metadata` with the plugins. In principle, if we have an image with metadata, we could even check the actual implementation of `get_keys` and so forth methods in a semi-isolated manner.

The integration tests unfortunately still require both marking and using the corresponding fixture, in case we test piexif / pyexiv2 explicitly as I couldn't figure out how to mark a fixture (instead of the tests) quickly. Could definitely be improved.

Let me know if you have any thoughts on improving this :blush:

Note that there are still two failing test cases. Is there a reason the `MetadataHandler` does not raise for `get_keys` and `get_metadata`?

karlch: Soo, you killed my commits? :joy: that is why I don't like pushing to other repositories ...

jcjgraf: Umm, that is not me force pushing. Let me investigate...

karlch: That's a bit creepy then, maybe some github "protection" mechanism, I will just push into the `core/metadataplugin-jcjgraf` branch and you can grab the commits :blush:

jcjgraf:
I am sorry that I caused you so much work. This look great, thank you very much.
Also, sorry for killing your commits. I use another git hosting platform for my work, and what I usually do is I let it mirror selected repos to GitHub. For some reason the mirroring system did not like you committing to my branch (maybe because you (apparently) force-pushed?) :confused: Anyways, I have disabled it for now and I have copied your commits over to this branch.

Here a few comments/questions to your changes:
- 200d123: Have you changed anything inside `class MetadataWidget`? Git is a little confused by the changed indentation (even when using `--word-diff`) and does not display *actual* changed properly.
- 5e37a61: Oh, great catch. Sorry for that silly mistake :see_no_evil:
- dd3750c: While in general, I think this approach is smart and reasonable, there is one thing I do not like about it. As-is, if one wants to use a 3rd party plugin (like the gexiv2 based plugin), but has piexif installed too, vimiv will use the piexif based plugin first, and only use the explicitly loaded one as a fallback. And I also think that it should be possible to disable metadata support if desired.
Maybe there is already a way to disable the automatic loading of `metadata`. If so, we would need to document it explicitly.
Has the `default` attribute assigned to `metadata` (and `print`) any impact?
- f2aff32: For once this is actually not a typo, but I meant *if and only if*. When using only *if*, `return True` would also be a valid body for a function with that docstring, would it not? :upsidedown: But for consistency reason it probably makes sense to change it :laughing:
- What are you plans concerning the output of `--version`/`:version`?

karlch:
No worries, thanks for investigating!

- I haven't changed anything in the widget itself, just changed the indentation which apparently created git havoc :laughing: 
- This is a valid point and I haven't found a valid solution, especially with internal metadata later-on. I'll go into more details later,  as I think this is the biggest remaining point, except for some cosmetics in the code.
- Oh, I see! However it is used like this all over the place, and I don't find it confusing, otherwise what is the point of the "if" in the documentation? So agreed, keep with "if" for consistency :blush: 
- Leave as-is for now, extend `:version` with plugin information in a separate attempt. I can tackle this after we have this merged :blush: 

### Metadata loading
Wanted features:
- Option to auto-import either pyexiv2 or piexif depending on what backend is available
- Option to enable / disable arbitrary backends
- Option to load multiple backends, especially considering internal keys, or other additional information
- Easy and understandable syntax. I am not a fan of going all havoc in the form of `metadata = one_of(pyexiv2, piexif), internal` or so.

I haven't found a valid solution. The easiest option could be to have `auto` as default, which would load one of pyexiv2 and piexif (current code) and to use whatever is defined if anything else is passed. E.g. `gexiv2, piexif` means load `gexiv2` **and** `piexif`. If you only want one, pass only one.

### Optional metadata
Currently `:metadata` is optional. However with internal keys, at least these will always be available, no? Or do we want to allow disabling the internal backend as-well? This would make the complete import havoc useless again :joy:

vimiv/gui/metadatawidget.py
karlch: Is currently used, may change later in case of internal keys, but we can stick to this for now IMHO. Feel free to re-open in case you think differently.

vimiv/imutils/metadata.py
jcjgraf: Changed in e16b7632a61cbf8b4e5617038f9d76836df7fd8e

vimiv/imutils/metadata.py
jcjgraf: Changed in f5d29bd4b6fe5a993dcd4f0a85232773abafa7fc

vimiv/imutils/metadata.py
jcjgraf: Fixed in 65c09598a828d705363e21cecbb562161b92041b

vimiv/imutils/metadata.py
jcjgraf: Fixed in 65c09598a828d705363e21cecbb562161b92041b

vimiv/plugins/metadata_piexif.py
jcjgraf: Done in 2835b92aa3a7394dd27328eea5f4309012ac51ed

vimiv/plugins/metadata_pyexiv2.py
jcjgraf: Done (see previous conversation)

vimiv/version.py
jcjgraf: No longer relevant, as *Leave as-is for now, extend :version with plugin information in a separate attempt. I can tackle this after we have this merged*

vimiv/plugins/metadata_pyexiv2.py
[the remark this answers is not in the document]
jcjgraf: Update on this; It is unfortunatelly not workin anymore. I get again the same error as before... :boom:

pytest.ini
jcjgraf: I think that has been resolved. I updated the PR description/todo to remember that we potentially need to improve the marking of fixtures

vimiv/gui/metadatawidget.py
jcjgraf: Updated PR Description/todo that we remember this

vimiv/version.py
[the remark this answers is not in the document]
jcjgraf: I have updated the PR description/todo that we remember this

vimiv/imutils/metadata.py
jcjgraf: I meant that the metadata is displayed in the widget in the order the keys are specified in the config. I have implemented that in dc830215b2aebf76d5666cf758c75f263dcabdc8

karlch: Makes sense, thanks :+1:

karlch:
I am honestly losing track a bit, may have gone a tad overboard with all the small sub-conversations :smile: 

Is this more or less ready and it is time for me to tackle the points you mention in the initial PR description, or is there more you would like to do before I take another look - and possibly break things horribly :joy:

jcjgraf:
I also had to fight hard to keep the overview :laughing: I have collected all open and (major) resolved TODOs in the description of this PR.

All code conversations should be resolved (or I have added them to the TODO list). So we can forget about all discussions before your comment from last week.

The code is not yet ready for another (/final) review, there are still a few things I need to do.

But we need to discuss how we do the loading of the plugins. If I get you correctly, you anticipate to load metadata plugins using the ``main'' metadata plugin? I agree with the *want features* you have listed. However, I think that the default plugin loading systems already provides **everything** we want, except for backward compatibility (i.e. pyexiv2 or piexif based support is enabled by default). TBH, I think the default plugin loading system is great, and I doubt that the overhead of *coming-up* with a new system, that provides the backward compatibility, is really worth it. But I would like to hear your opinion on this.

In addition, I need you opinion on this: *Should keys be case-insensitive? Currently they are not.* Personall,y, I think both is reasonable

karlch:
Thanks for the update! I will not be able to look into this this week, hopefully next week on another train mission :smile: 

Concerning the loading, I don't see why this couldn't be both backwards compatible and use the current structure. We could simplify everything I mentioned to have something like the following:
- Rename the current `metadata` to `metadata_compat` and activate it by default. This only does what it currently does, and throws a deprecation warning asking the user to configure one of the two (or their own).
- Keep everything else as you suggested and use the plugin-system, no other magic in the init with parsing stuff :blush: 

I don't have an opinion on it being case-sensitive or not, this is up to you. Maybe you have experience how others do it? If case-sensitive is at least sometimes used, I would stick to it and keep things simpler.

jcjgraf:
No worries, take your time!

This sounds good to me. But, how can one disable `metadata_compat`? If it is not possible to disable it, then vimiv will favour pyexiv2/piexif over any other plugin that the user loads.

I guess I will leave it case insensitive in this case.

BTW, feel free to commit your changes directly to this branch. I do not think it is worth opening a disussion for each typo I have done :see_no_evil: (And if my other remote is stubborn again, just work on a derived branch :blush: )

karlch:
Sorry for the further delays in the schedule, currently really having difficulties fitting this in. I will get to it, and definitely still have this on my mind, but don't see it happening this week :disappointed: 

I have some ideas that I would like to try out, e.g. loading user plugins before the app ones, then `metadata_compat` could check if another `metadata` plugin has been loaded, and react accordingly.

Will try to commit here directly then :blush:

jcjgraf:
I have rebased this branch for master, and updated and restructured the TODOs in the PRs description. The restructuring should make it better visible what has been done in this PR and what is left to do. This should (hopefully) facility the review of this PR [1] 😇

In addition, I have added a proposition (fb66610) to the last outstanding issue:
For backward compatibility, we want that a metadata backend is automatically loaded. This was done by the `metadata` plugin. However, this approach had the issue, that either `metadata_piexif` or `metadata_pyexiv2` was always loaded, even if one explicitly loaded a different metadata plugin. I have changed that such that a default plugin is only loaded when no other metadata plugin has been loaded. Using this approach, it is still not possible to use vimiv without having any metadata plugin loaded (in case piexif or exiv2 is installed). Though, it is not that big of an issue I think. Especially should be decide to deprecate having metadata support by default.

[1] Since you have been a bit more active last week, I had the hope that you may have a look at (and review 🫣)  this PR when get it up to date again. But no pressure - take the time you need 😬

karlch (approved):

karlch:
So, I was finally able to give this another try. Excellent work, I believe almost all issues are solved and this is ready to be merged.

I tried to tackle some of the last bits in the `metadataplugin` branch here directly for you to check (feedback / comments welcome as always) and cherry-pick / adapt accordingly:
- Plugin loading: I slightly refactored this so `metadata` can deal with this directly and must not be handled specially. We can also disable autoloading by passing `metadata = none` in the configfile.
- Tests: marks are now registered automatically once `piexif` or `pyexiv2` fixtures are included. Could probably be extended beyond this single test file at some point in the future.
- CI: I would like to get at least one runner with `pyexiv2` back in. The simple adding died, so please ignore this for now, and I will continue debugging after merging.

Final remaining part would be the documentation points you mentioned.

jcjgraf:
Thank you very much for taking time to review this PR!

Your changes in the `metadataplugin` branch look good to me and I have cherry-picked them over. The CI stuff I do not get, so here I gotta fully trust you there😁 So we keep `pyexiv2` in the CI for the future?

I have updated the docs. Let me know if there is anything to improve. Otherwise, I am done with this PR 😃

karlch:
Thanks a bunch for the documentation, I will either just merge or push changes here directly once I get to it, no more need for another iteration I hope :tada: 

CI: would you mind removing that one commit again? This currently doesn't work unfortunately, thus the "please ignore this for now" :wink:

jcjgraf:
Oh, I though I should ignore that it is broken 🙈 I have remove it

Yeah, if there are not major things, feel free to correct everything yourself directly.

karlch:
Merged - so ... party? :smile: :tada:

jcjgraf:
Well, that is simply amazing🎉

Metadata support has been a burden for such a long time, and how we have finally sorted it out for once and for all [1]

[1] Better do not quote me on this 🫣 :laughing:

PR #478, opened 2022-03-31
Possibility to Display Command stdout

jcjgraf:
When appending a `>` to a external command, display the stdout in vimiv.

This is just a proposition and up for discussion

#### TODO
- [x] Implementation
- [ ] Changelog
- [ ] Doc

### DIFF
--- a/vimiv/commands/external.py
+++ b/vimiv/commands/external.py
@@ -49,10 +49,11 @@ class ExternalRunner:
     def run(self, text: str):
         """Run an external command text."""
         text = escape_glob(text.strip())
+        stdout = text.endswith(">")
         pipe = text.endswith("|")
-        split = shlex.split(text.rstrip("|"))
+        split = shlex.split(text.rstrip("|").rstrip(">"))
         command, args = split[0], split[1:]
-        self(command, *args, pipe=pipe)
+        self(command, *args, stdout=stdout, pipe=pipe)
 
     @api.commands.register()
     def spawn(self, command: List[str], shell: str = "sh", shellarg: str = "-c"):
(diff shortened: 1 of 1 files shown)


jcjgraf:
I agree with you that the syntax is weird. I was inspired by the `|`, when I added the stdout functionality. 😁

What about this idea:
- Deprecate `|` and instead add a flag to `:open`, which interprets the argument as a shell expression, and opens the result file path list (as with `|`).
- Add a new command `:show-stdout` which toggles a widget where the output of the last command mode command is shown
    - This is somewhat similar to vim's `g<` shortcut (which is syntactically also a bit weird)
    - Instead of only the output of the last command, we could also show the output of the last $n$ commands or something
    - We could maybe also differentiate between stdout and stderr (using multiple commands or flags)
    - We could also add a way to display the exit code (maybe even wrap it as a statusbar module?)
   
This may be not the simplest solution to implement, but for me it seems like the most user friendly one.

karlch:
I like parts of this a lot, but have a small modification in mind. For the displaying, how about a setting instead, with the following options:
- `always`: the widget is always open and displays the last command. I honestly would never use this option, but don't see harm.
- `auto`: the default? show the widget if a command has stdout and hide it once any new (vimiv) command was run. This gives the user an indefinite amount of time to read stdout, but does not clutter the gui permanently.
- `never`: obvious I guess.

I agree with showing both the output and the returncode in the widget, but probably wouldn't keep `n` outputs. This seems a little over the top to me. We could of course add a signal to the `api` that emits the returncode and the output, so anyone can do absolutely anything with it in a plugin.

Concerning the flag to `:open`, is it really worth cluttering this already involved command more for the sake of saving a new command? I suppose an actual independent command would actually be both easier and more discoverable, don't have a name in mind though. Deprecating `|` definitely seems the way to go no matter what.

jcjgraf:
I like the idea of having an option that controls the general behaviour of the output widget. Nevertheless, I think it would be really useful to have a command that toggles the widget. The reasoning is the following; I import my photography from my CD cards to my archive using a shell command. I run that command from within vimiv, after marking all images to be imported. This is a rather longish-running command, and therefore, it would be really useful to have a look at the progress/status of the import.

It is up to you to decide whether to add a new command or use a flag. Both makes sense and is fine with me.

karlch:
You are right, a dedicated command similar to `:metadata` makes a lot of sense.

When hearing toggle, I immediately thought of a boolean setting, but somehow added the `auto` thingy. If we would ever want this in addition, we could still add a setting for `autohide`, but this should be separate of the command and not mingled.

I would go for a new command then, just to keep `:open` a bit less cluttered :blush:

PR #491, opened 2022-04-21
chore(ci): add informational Codecov status checks

thomasrockhu-codecov:
Hi, Tom from Codecov here. I noticed that you were using Codecov but weren't actually getting any notifications on pull requests. I figured it would be useful to get some idea if code being changed is being tested, but also not blocking CI/merging. Let me know if this makes sense or if we can do something that would be helpful.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,7 +1,11 @@
 coverage:
   status:
-    project: off
-    patch: off
+    project:
+      default:
+        informational: true
+    patch:
+      default:
+        informational: true
     changes: off
 
 comment: off


karlch:
(almost a year later .. sorry and Happy New Year! :tada:)

Thanks for the suggestion. Previously, I have found the codecov comments on PRs to be very verbose and rather distracting, especially if there are many small PRs with just a few lines of code (or even the automated dependency ones). But you are of course right that this would make sense to use. I will give it a shot for a while, and disable it again if it still seems too noisy for me :blush:

karlch:
Thanks for checking back, this has indeed been useful and unobtrusive, thanks a lot!

PR #563, opened 2023-01-07
Support for different sorting options for images and directories

### LINKED ISSUE #75: Different image sorting than by name

### LINKED ISSUE #386: [Question] How to sort the library?
In my current directory I have files numbered from 1 to 16. If I open vimiv and than the library by pressing `tl` I'm getting this:

image

How can I sort the names from 1 to 16 instead of `1.jpg` than `10.jpg` (instead of `2.jpg`)? I've tried to type `:sort` but this command doesn't exist.

karlch:
Adds support for sorting images and directories in different ways via the `image_order` and `directory_order` settings.

fixes #75
fixes #386

Supersedes and builds upon the work done in #387 and #76, please refer to these PRs and the linked issues for further detail.

@jcjgraf getting there, I will take another look tomorrow .. err .. later today and try to add a few more tests for the filesize / mtime based sortings. Minor changes compared to your implementation are:
- Do not completely reload the working directory, as the files themselves haven't changed, only reorder.
- Perform the actual sorting in the setting, to avoid having to pass around the tuple of the previous `get_para` method. This goes hand in hand with the removal of sorting in `files.listdir` as described in one extra comment I left on your PR.

Thanks a bunch for the work you put into this, was super pleasant to handle the code and pick it up :blush: I didn't squash the commits as I do like seeing the thought-process and don't mind it being messy, it's coding afterall :joy: Let me know if you have any thoughts, also concerning the new `case-insensitive` ordering (not sure about the name here).


vimiv/api/settings.py
```diff
@@ -315,8 +316,51 @@ def __str__(self) -> str:
...
+    """Stores an ordering setting."""
+
+    typ = str
+
+    ORDER_TYPES: Dict[str, Tuple[Callable[..., Any], bool]] = {
+        "alphabetical": (str, False),
+        "alphabetical-inverse": (str, True),
+        "natural": (natural_sort, False),
+        "natural-inverse": (natural_sort, True),
+        "case-insensitive": (lambda s: s.lower(), False),
```
jcjgraf, line 329: `str.lower` is sufficient, there is no need for a lambda function.

jcjgraf:
Great progress! And thanks for taking over. Here a few comments of mine:

1. I like the slight refactor and putting the sort function directly into `OrderSetting`.
2. I have found an inconsistently in the order of images. Namely, when opening multiple images (of potentially different directories) using the `:open` command, the images are not order according to the specified ordering in the thumbnail view. I can see to two issues:
    a) We order the images in the CWD and not the ones in the filelist themselves. Sorting the `paths` in `imutils/filelist/_set_paths` seems to resolve the issue, though I am not sure if this is the proper way to fix it.
    b) The order key functions must consider only the image names instead of the absolute paths. I.e. instead of changing `alphabetical` to using absolute paths (c.f. aca4e9e3f39e7c5210dbcbf9dd5b28641afff45c), we need to adapt `natural_sort` to also only consider the image names only.
3. Having the option `image_order`, the option `shuffle` does not really make any sense anymore. Instead, I opt for having a new sort order named `random`, that sorts images/directories according to a random permutation.
4. Having a case insensitive ordering makes sense. However, insensitive sorting is actually rather an *flavor* of sorting `alphabetical` and `natural`. I.e. `case-insensitive` is a version on `alphabetical`, and therefore, there should also be a way to sort naturally, ignoring cases. It is a little bit verbose, but what about having sort types:
    - `alphabetical_case-sensitive`
    - `alphabetical_case-insensitive`
    - `alphabetical_case-sensitive_inverse`
    - `alphabetical_case-insensitive_inverse`
    - `natural_case-sensitive`
    - `natural_case-insensitive`
    - `natural_case-sensitive_inverse`
    - `natural_case-insensitive_inverse`
5. Currently, directories are always displayed before images. While already at it, it may be beneficial to have a way to make directories and images *mix*. What do you reckon? I have not looked into it and I do not know how involved this would be to implement. So, personally, I am also fine with leaving it away if it is a too much of a hassle.

Let me know what you think concerning my point. I hope it does not cause too much additional work for you :see_no_evil:  If there is anything I can help you out with, let me know.

vimiv/api/settings.py
karlch: Nice! Will apply. Never seen it used that way though :smile: i.e. always `text.lower()` not `str.lower(text)`

karlch:
Thanks for your comments :blush: 

2. That's unfortunate, I will look into this, thanks for catching!
3. Not quite sure about this one. Does shuffling stuff in the library make sense? The nice thing about shuffle (to me) was that it only applied to the image list, and I could still easily find what I like :thinking: 
4. Agreed, I actually thought about this briefly and then threw it away. Instead of 4 new setting options, 8 total, how about a new setting for case sensitivity? We could even group them into a new group then, i.e. `sort.image_order`, `sort.directory_order` and `sort.ignore_case` :thinking: 
5. This is a valid point, but I would root for having this as a new issue after this PR, just to get this done in the first place. Not sure how much extra complexity this adds.

jcjgraf:
3. I get your point. It is just that both settings do similar things (change the order of images), which may make it more confusing for users :thinking: 
4. This sounds good to me. The only issue I see is that `ignore_case` has no effect if neither `alphabetical` nor `natural` is selected, which, again, make things a bit more confusing. But otherwise, I think that is the cleaner option. Grouping the ordering into a `sort` group makes anyways a lot of sense. If we leave the shuffle as is, I would argue that it may be a good idea to put it into the `sort` group too.
5. 👍

jcjgraf:
3. 👍
4. I did not even think about that, but it makes a lot of sense. Go for it :laughing: :

karlch:
So I tried to implement this, especially sorting the image stuff makes things a bit more intertwingled than I had hoped for. Would you mind giving this another check? I would try to update the documentation tomorrow then, do any final changes, and merge :blush:

One thing I am not sure about (but that wasn't really officially supported anyway) is that this now re-orders user-input in case of e.g.
```sh
vimiv b.jpg a.jpg
```
and the same when calling
```
:open b.jpg a.jpg
```
Both are opened in order `a.jpg`, `b.jpg`.

This doesn't feel right, but would require another branch of logic, in case we want to separate from e.g.
```
:open *
```
and there is no way we can know if the user passed more than one arguments to the shell, or used some type of shell glob expecting us to sort ...

jcjgraf:
c06c5e63868bf205de9786f5b0fbc6988fbc3441 fixes the issue. Also the other commits look good to me.

> [...] this now re-orders user-input in case of e.g.

I do not really see this as an issue, as, personally, I would not expected that the images are sorted in the provided order.

The only thing I have left to criticize is that changing `sort.image_order` has direct effect on the images in the filelist (i.g. images get reordered), but `sort.reverse` and `sort.shuffle` only take effect after the filelist is recreated. I am not sure why there is this discrepancy between behaviour of changing these settings. Do you see why this is the case?

karlch:
Good catch, shuffling should now also immediately affect the image filelist. Reverse already did before for me, or do you mean something else?

Maybe not in the same order, but possibly at least the first path should be the one opened :thinking:

jcjgraf:
Perfect, not everything works as expected.

I do not really have a strong option on that. I would say we can leave it as is, and if it should be a problem sometime in the future, we can still think of a workaround.

jcjgraf:
Congrats on merging! I am glad we can cross something from our todo list :partying_face:

karlch:
Same here :tada: and thanks again for all your effort that went into this! :blush:

Issue #525, opened 2022-08-07
[Feature Request] PDF Support


mcp292:
It would be nice to use one image viewer, with one set of bindings and behaviors, to view both images and PDFs. I think `vimiv` has a great opportunity to be this tool. Especially since `mupdf` lacks configurable keybindings.

jcjgraf:
Hi @mcp292!

For multiple reasons I do not think that makes sense to have an image viewer (i.e. vimiv) which is also a fully blown pdf viewer. While they certainly share some aspects, like browse through image/pdf collection, zoom in and pan, there are noticeable difference like, for image you want EXIF support, basic editing, while for PDFs you want search functionality, annotation, etc.

However, if we are talking about **really** basic PDF support, i.e. simply display PDF as it was an image, I can see its potential. Though, I am not sure if this belongs natively into vimiv or should be provided via a plugin. As you may have noted, vimiv if very flexible and provides a powerful plugin system.

The API interface (`api.add_external_handler`) is used by plugins to make Vimiv capable of handling image formats, which are natively not supported (see [this](https://github.com/jcjgraf/RawPrev) plugin which adds RAW support). It would be (probably) rather straight forward to display PDFs this way. You could use e.g. [ImageMagick](https://stackoverflow.com/a/6605085/5464989) to convert PDFs to JPGS on-the-fly.

I do not know what the "indented" way is do display PDFs in QT. In PDF viewer application, which I quickly had a glimpse on, PDFs were displayed using the `QWebEngineView`. Vimiv-internally, images are displayed as a `QPixmap`. We have also made some exceptions to this, namely for SVGs and (automated) Gifs (see `gui/image/ScrollableImage`). If PDF  support is added in a similar simple and clean way, it may be something which we could consider. More "advanced" features could still be implemented as a plugin to prevent unnecessary clutter.

I think having basic PDF support, either implemented as a standalone plugin, or with minimal changes to the core, would be nice to have. But I do not see the point of implementing advanced featured in vimivs core (but, in a plugin one can do whatever one likes :) ).

What would you personally expect from "PDF support"? 

I am rather busy and not able to look into this, and unfortunately, @karlch is also super busy. But if you would like to look into this, I could assist you.

mcp292:
I agree with you. Really just looking to open up and scroll through a PDF with bindings I'm used to. With basic functionality, it could even be used for PDF slideshows. For anything more serious, using a dedicated PDF viewer or adding a plugin is fine. 

[...]

PR #575, opened 2023-01-08
Support for PyQt6 (and PySide6)

### LINKED ISSUE #318: Support for (Py)Qt 6
As both Qt6 and PyQt6 have been released, we should think about how to support them. I suppose we will need to support both (Py)Qt5 and (Py)Qt6 for a while and then at some point in the probably distant future drop support for version 5. Therefore this will involve:

- [ ] adding our own wrapper module around PyQt and using it everywhere instead of importing PyQt5 directly
- [ ] changing the existing usage of (Py-)Qt-related code where needed
   - [ ] only use scoped enums, Florian Bruhin wrote an excellent script for this over here https://github.com/qutebrowser/qutebrowser/issues/5904
   - [ ] switch to `QRegularExpression` wherever we use `QRegExp`
   - [ ] removing the trailing `_` in `exec_` and `print_` calls
- [ ] figuring out which is the minimum required version of PyQt5 after these changes
   
and probably a lot more things I cannot think of at the moment :blush:

karlch:
Initial attempt at supporting PyQt6 and potentially PySide6 by writing a Qt wrapper module `vimiv.qt` which must be used instead of importing from PyQt* / PySide6 directly.

This went surprisingly smooth, but still has a few rough edges, and some are breaking before being able to merge:

PyQt6:
- [ ] Update documentation
- [x] Fix pylint `no-member` errors
- [x] Fix mypy

PySide6 additionally:
- [ ] Fix end2end tests and add to CI

I also noticed performance issues upon startup with PyQt6 on my local laptop. These do not seem to affect the CI speed, and are also not relevant after startup, which puzzles me. Namely, the instantiation of any QWidget seems to be really slow, `__init__` of the main window takes several hundred ms instead of less than 30 with another wrapper.

If anyone, @jcjgraf :wink:, would like to test this, that would be super helpful.

Going forward, I believe it makes sense to merge this after releasing `v0.9.0`, but explicitly calling support for PyQt6 and especially PySide6 experimental. If we don't merge soon-ish, there will continue to be a bunch of merge conflicts for obvious reasons. This would probably mean we need a `v0.10.0` with this in, before going to `v1.0`. However, we could still release both of these versions with PySide6 if issues are too large, but PySide6 is the "official" Qt for python now, and the performance issues are not there for me.

fixes #318


jcjgraf:
Wow, you are amazing. I hope there is some kind of tool that helped you with the transition and that you did you have to to everything by hand :see_no_evil:

My general thoughts on the transition to PyQt6 (I have not yet had a close look at your code and the changes); Is there a strong necessity for supporting both version at the same time? I.e. are there some distros out there that do not support PyQt6 yet? Or are there some unforeseen issues we may get with PyQt6? Or do we want to maintain PyQt5 support only for people who do not yet want to update for some personal reason? While having a smooth transition with a transition period from PyQt5 to PyQt6 would be favorable, I am thinking about the economy of this approach. If it induces a multiple-hours overhead, compared to directly updating, I would argue that, having the limited manpower we have, it is not really feasible. 

Concerning the performance issues; What kind of profiling tool did you use?

I will certainly look into it, but not right now... First I would like to get #467 done :blush:

karlch:
This has been done before, e.g. by qutebrowser, so I didn't have to do too much thinking :joy: 

I do think it makes sense to have the wrapper module in any case, no matter what we end up supporting, as we can e.g. also use it for PyQt6 vs. PySide6 and it really helps with experimenting. And, given most of it just works, I would leave it in until PyQt5 is basically dead. It probably makes sense to drop support for PyQt < 5.15 rather soon though, as those have some extra issues that require manual workarounds.

The super-fancy `vimiv.utils.debug` module with the timed decorator / profile context manager :smile: nothing crazy, but as the difference was > 10x, and noticeable by feeling, I don't think the details really matter.

Thanks, and no rush, the other open PRs are definitely more important atm :blush:

karlch:
Since mypy, lint and tests now do what they should, the PyQt5 / PyQt6 part is now essentially merge-able and definitely ready for review :tada:.

I disabled auto-detecting PySide6 for now as there is a lot more work to do.

As this is always a hassle to rebase (will probably just merge `master` into here or do a final merge), I would prefer to have this in sooner rather than later, but not in the already extremely blown up `v0.9.0`. However, I think we can release `v0.9.0` after #650 is in, what do you think @jcjgraf ?

After the release, I would like to prioritize getting this into `master`.

In case bugs come up, we can always release a follow up `v0.9.X` from the `v0.9.0` tag. One example would be #652, which may end up being quite some effort.

karlch:
Another note: replacing the `api.objreg.register` decorator with a regular call within the `__init__` fixes the performance issues with `PyQt6`. Absolutely no idea why, and that is quite unfortunate, as it is pretty deep in the `api` and also used by plugins.

Details:
```python
# slow
@api.objreg.register
def __init__(self):
    ...

# fast
def __init__(self):
    api.objreg.register(self)
```

As this is a pretty major issue, I think changing the behaviour, while keeping backward compatibility and logging a deprecation warning, is fine.

vimiv/qt/__init__.py
```diff
@@ -0,0 +1,76 @@
...
+_WRAPPERS_PYQT = [WRAPPERS.PyQt5, WRAPPERS.PyQt6]
+_WRAPPER_NAMES = [wrapper.value for wrapper in WRAPPERS]
+_WRAPPER_NAMES_STR = ", ".join(_WRAPPER_NAMES)
+
+WRAPPER = None
+
+
+def _select_wrapper():
+    """Select a python Qt wrapper version based on environment variable and fallback."""
+    env_var = "VIMIV_QT_WRAPPER"
```
jcjgraf, line 35: Why do you introduce a custom env variable instead of relying on the *standardized* `QT_SELECT=5/6`?

jcjgraf:
Thanks a lot for your hard work on this. I had a glimpse at the new QT module and that makes sense so far. To me it seems like there are no major changes between QT5 and QT6, except a few methods renames. Or did I miss anything?

The rough roadmap makes totally sense to me. Everything for `v0.9.0` is done, except for #325. But I do not think that is relevant for that milestone anymore. So I guess we can make a new release, and after that focus on getting QT6 in.

The issue with `api.objreg.register` is really unfortunate. But at least you figured out that this is the cause of the performance degradation. Maybe we can find a way to mitigate/fix the issue while still keeping the decorator.

vimiv/qt/__init__.py
karlch: How would we differentiate between `PyQt6` and `PySide6`?

jcjgraf: Just take whichever is available. I doubt that users have both installed. And if so, why should they care which one is used?

karlch:
Thanks for checking! :blush: Yes, except for a few renames / deprecations, and the annoying new enum access, nothing major implementation-wise.

Sounds good :+1: I will try to release the new version over the week-end then.

I see two (maybe three) options:
1. Deprecate the decorator and use as described above.
2. Go back to the "old-style" `objreg` where we store everything in a dictionary again (`objreg._registry[cls] = component`) instead of directly in the class via `cls.instance = component`.

Personally I am leaning towards 1. as it actually also makes the tests a bit cleaner (no need to hack-mock decorators). But from a "plugin-writer" perspective 2. is definitely easier, sure.

vimiv/qt/__init__.py
karlch: I care for testing both without changing the source code all the time :smile: I do think it is nice to expose this option in any case, but we could split it into `QT_SELECT` and `USE_PYSIDE`, if you think this has some benefit? Personally, I find one easier than two though.

jcjgraf: True, for development/testing/debugging this would certainly make sense to explicitly choose between QT and Side. I just think that if there is a standard, it would make sense to follow it. What about having only `QT_SELECT` but allow not only the values 5/6 but also the explicit options PyQt6/PySide6?

jcjgraf:
> I will try to release the new version over the week-end then.

Awesome, let me know if I can help with anything!

> I see two (maybe three) options

I need to have a close look at it and investigate a bit. Do you know how other projects deal with that?

vimiv/qt/__init__.py
karlch: I like this compromise :+1: So our `QT_SELECT` would then have:
* 5 -> PyQt5
* 6 -> PyQt6 or PySide6 with a preference that we set
* all three explicit options as currently done by `VIMIV_QT_WRAPPER`

karlch:
Release is out, haven't heard any complaints yet :smile: 

I remember qutebrowser having an `objreg`, probably with string names though instead of our version with one per-class. Our main "issue" is that we register commands using the decorator during a time when there are no instances. I really like this syntax though as you can put the command registry right where the function is defined ...

The way it is usually done is probably something along:
- Always have the instances where you need them code-wise
- Have a central place to register commands / keybindings

jcjgraf:
Thanks and congrats for a new release! That one took quite some time 😁

I have rebased the branch for the current master [in my fork](https://github.com/jcjgraf/vimiv-qt/tree/qt6). I have also changed the env variable to `QT_SELECT` as discussed above. I won't have any further time to work on this this week. In case you have time, you may continue where I left off 😉

karlch:
Thanks, I will take a look once I find time for more changes.

However, I have also merged some stuff locally without pushing into another branch and played around with the `objreg`, so not sure if it is easier for me to continue off your rebase (except for picking some `QT_SELECT` parts).

jcjgraf:
> except for picking some QT_SELECT parts

I have also created a testcase for it (same commit). I am not sure if you have seen that. Nothing amazing, but better then nothing, I guess.

> I remember qutebrowser having an objreg, probably with string names though instead of our version with one per-class.

Right, but they [do not seem to be completely happy](https://github.com/qutebrowser/qutebrowser/issues/640) abut it (not related to our issue though...).

I have investigated into the objreg weirdness. I have condensed the issue down to the following MVE:

```python
import time

from PyQt5.QtCore import QObject as QObject5
from PyQt6.QtCore import QObject as QObject6


def timed(func):
    def inner(*args, **kwargs):
        start = time.time()
        return_value = func(*args, **kwargs)
        elapsed_in_ms = (time.time() - start) * 1000
        print(f"{func.__qualname__}: took {elapsed_in_ms:.3f} ms")
        return return_value

    return inner


def register_decorator(func):
    def inner(component) -> None:
        cls = component.__class__
        cls.instance = component
        func(component)
    return inner


def register_function(component):
    cls = component.__class__
    cls.instance = component


class DecoQt5(QObject5):
    @timed
    @register_decorator
    def __init__(self) -> None:
        super().__init__()
        print("QT5 decorator")


class FuncQt5(QObject5):
    @timed
    def __init__(self) -> None:
        super().__init__()
        print("QT5 function")
        register_function(self)


class DecoQt6(QObject6):
    @timed
    @register_decorator
    def __init__(self) -> None:
        super().__init__()
        print("QT6 decorator")


class FuncQt6(QObject6):
    @timed
    def __init__(self) -> None:
        super().__init__()
        print("QT6 function")
        register_function(self)


a = DecoQt5()
b = FuncQt5()
c = DecoQt6()
d = FuncQt6()

assert hasattr(DecoQt5, "instance")
assert hasattr(FuncQt5, "instance")
assert hasattr(DecoQt6, "instance")
assert hasattr(FuncQt6, "instance")
```

The big difference between the `register_decorator` (i.e. the old register decorator) and `register_function` (i.e. the new register function) is that the former stores the `component` before the superclass (`QObject` in this case) is initialized, while the later does it after the initialisation. For `FuncQt6`, if one swaps the `super().__init__()` and `register_function(self)` in there is a delay too.

So, our issue is not the decorator, but the execution order. I think we can just move the `func(component)` before the assignment to `cls.instance` in the decorator and we are good, aren't we? Like:

```diff
diff --git a/vimiv/api/objreg.py b/vimiv/api/objreg.py
index cb974c3a..8dc350b4 100644
--- a/vimiv/api/objreg.py
+++ b/vimiv/api/objreg.py
@@ -57,10 +57,10 @@ def register(component_init: Callable) -> Callable:
         Args:
             component: Corresponds to self.
         """
+        component_init(component, *args, **kwargs)
         cls = component.__class__
         _logger.debug("Registering '%s.%s'", cls.__module__, cls.__qualname__)
         cls.instance = component
-        component_init(component, *args, **kwargs)

     return inside
```

> New "issue" is that the PyQt6 install actually seems to "support" pdfs. I tested this with a quick function, and we do get a preview of the first page as discussed ages ago in #525. I would add pdfs to the ignore part for now, and re-consider this in a new PR if this is fine with you, as I am not sure what we want here.

Which part of QT is supporting PDF? I only found [this](https://doc.qt.io/qtforpython-6/PySide6/QtPdf/index.html), but that is specific to PySide6. As long as the they are not detected by `utils.imageheader` as valid images, they are not further considered. So, why explicitly ignoring them?
Anyways, I would certainly not consider PDF support in this PR.

karlch:
I did see it, but the comment "The test case is failing, as it should skip tests, based on installed QT binding." made me think this is a problem for future us :smile: 

Yep, I actually remember seeing this issue ages ago and switching from names to types, to at some point actually just having `.instance` which is pretty close to storing one in the module and using `.get()`. We have it a bit easier, as we don't need multiple instances of the same class.

That is an excellent catch, thanks for digging!! :blush: I reverted the function commit and simply changed the order in the `qt6-merge` branch.

What I mean with "supports" is that `QImageReader.supportedImageFormats` actually includes pdf and fails the tests, i.e.
```
        """Tests if all formats readable by QT are also detected."""
>       assert name in NOT_DETECT_FORMATS or name in DETECT_FORMATS, "Missing check"
E       AssertionError: Missing check
E       assert ('pdf' in ['svgz', 'wbmp'] or 'pdf' in ['jpg', 'png', 'gif', 'jp2', 'webp', 'tiff', ...])
```
Fixed by adding the pdf into the `NOT_DETECT_FORMATS` in 05131323cf564c9476dc6c2046021c51d435d22f, this is what I meant with ignore.

jcjgraf:
> What I mean with "supports" is that QImageReader.supportedImageFormats actually includes pdf and fails the tests

That is weird, I checked their [docs](https://doc.qt.io/qtforpython-6/PySide6/QtGui/QImageReader.html#PySide6.QtGui.PySide6.QtGui.QImageReader.supportedImageFormats), and PDFs should not be supported. Have you installed any PDF QT module? Does it also happen in the CI or only locally?

Do you know what is going wrong here? I have regenerated the .venv.

```
$ tox -e mypy                                                                                          12:40:46
.pkg: _optional_hooks> python /usr/lib/python3.11/site-packages/pyproject_api/_backend.py True setuptools.build_meta __legacy__
.pkg: get_requires_for_build_editable> python /usr/lib/python3.11/site-packages/pyproject_api/_backend.py True setuptools.build_meta __legacy__
.pkg: get_requires_for_build_sdist> python /usr/lib/python3.11/site-packages/pyproject_api/_backend.py True setuptools.build_meta __legacy__
.pkg: build_wheel> python /usr/lib/python3.11/site-packages/pyproject_api/_backend.py True setuptools.build_meta __legacy__
.pkg: build_sdist> python /usr/lib/python3.11/site-packages/pyproject_api/_backend.py True setuptools.build_meta __legacy__
mypy: install_package> python -I -m pip install --force-reinstall --no-deps /home/jeanclaude/Documents/Programming/vimiv-qt/.tox/.tmp/package/210/vimiv-0.9.0.tar.gz
mypy: commands_pre[0]> .tox/mypy/bin/python scripts/maybe_build_cextension.py
mypy: commands[0]> pytest
Error importing a valid Qt python wrapper:
No valid Qt wrapper found. Valid options: PyQt5, PyQt6, PySide6

Please install / configure a valid wrapper to run vimiv.
mypy: exit 2 (0.56 seconds) /home/jeanclaude/Documents/Programming/vimiv-qt> pytest pid=20083
.pkg: _exit> python /usr/lib/python3.11/site-packages/pyproject_api/_backend.py True setuptools.build_meta __legacy__
  mypy: FAIL code 2 (5.12=setup[4.50]+cmd[0.06,0.56] seconds)
  evaluation failed :( (5.32 seconds)
  ```

karlch:
Both in CI and locally, I mean, `tox` runs a virtualenv anyway. An example action is [this one](https://github.com/karlch/vimiv-qt/actions/runs/5661853076/job/15340611318).

Yep, `mypy` now needs a flag, could probably consider a default, but currently only `mypy-pyqt5` or `mypy-pqt6` run.

jcjgraf:
> Yep, mypy now needs a flag

Ah sure, silly me 🙈

> `QImageReader.supportedImageFormats` actually includes pdf

True, same for me.

Is there anything else I can help with concerning this PR?

karlch:
Perfect, not really, in case you have no further comments, I would go ahead and finalise documentation and merge into master :blush:

jcjgraf:
No, from my side everything is good :+1:

Issue #579, opened 2023-01-09
Deprecation of imghdr module in python 3.11


karlch:
Starting from python 3.11 the `imghdr` module, which we use to check if files are images, is deprecated and will be fully removed starting from python 3.13.

I see three options moving forward:
1. Copy `imghdr.py` to `vimiv.utils`, adapt slightly, and continue using it.
2. Use a new magic library. The only one recommended in the deprecation [pep](https://peps.python.org/pep-0594/#imghdr) which seems usable for us is [filetype](https://pypi.org/project/filetype/). Simply using `filetype.is_image(...)` is not enough though, as we don't support all image formats that filetype recognizes. This is possibly also true for `imghdr`, but not as critical as these are rather obscure. filetype however also recognizes e.g. `xcf` files from gimp. So we would need to slightly update for which filetypes to check and use `filetype.match(path, matchers=CUSTOM)`.
3. Switch to mimetype-based type inference instead. I don't like this option personally and only mention it for completeness as we could just use `QtCore.QMimeDatabase` then.

I don't have a strong preference between 1. and 2. yet. Copying `imghdr` is probably straightforward and doesn't require much `vimiv` code, but puts the maintenance burden on us. `filetype` seems reasonably well packaged, the code structure is a bit weird at places (imports get overwritten all over the place, especially `filetype.types`), and could possibly allow us to remove any custom checks (if upstream supports everything). Should probably try it out and see how it goes, maybe communicating some issues upstream.

jcjgraf:
This is unfortunate. While it introduces another dependency, to me it seems like using filetype is the way to go.

If you like, I can have a lot at it and propose something. But unfortunate not earlier than in a month.

karlch:
I was honestly leaning the other way, given copy-pasting is nearly effortless, and the code there seems to have worked for ages :thinking:

jcjgraf:
You are right. I had a first look at the code of imghdr, and it is indeed pretty much as simple as it gets. I just thought that since there is a well maintained package which does what we are looking for, why not using it?

karlch:
@buzzingwire

The reason I proposed `filetype` instead of `python-magic` was the performance. This is a really dirty script that checks an image and not an image 10k times for the result and prints the time in seconds (using `timeit`):
```
# filetype
0.18984456099860836
0.15069496299838647
# magic
1.231587242000387
0.1917692560091382
# puremagic
4.894981782999821
1.919188183004735
# imghdr
0.11998687899904326
[...]

PR #646, opened 2023-06-24
WIP: Internalize deprecated imghdr packet

jcjgraf:
This PR internalizes the [code from imghdr](https://github.com/python/cpython/blob/3.11/Lib/imghdr.py) into our our code base. The code newly resides in `utils/imghdr`. The process is fairly straightforward and also allows for simplifications in `utils/files`, as our own custom types can be moved to the other imghdr code.

Steps:
- [x] Copy imghdr code over
- [x] Cleanup of copied and own code
- [ ] Copyright header
    - imghdr is licensed under "Python Software Foundation License Version 2", which is GPL-compatible, but does not allow re-licensing.
- [ ] Should we rename `imghdr.py` to `imageheader.py` or `imagetype.py` or `imagedetection.py` or something?
- [ ] How to deal with the `imageformats` plugin?
- [ ] Testing
    - For that we would need a series of "real" image files of different type. Or what?
- [ ] Docs

implements #579


karlch:
Thanks a bunch, this is excellent work! :blush: I love that you removed the files not supported by Qt, that `imghdr` would have caught as well, and restructured to simplify things significantly.

Let me see
* Copyright: this is a tricky one where I have **zero** expertise ... Would it be enough to add another comment stating that this file specifically was taken from python and is licensed differently, i.e. under "Python Software License Version 2"?
* Rename: sure, why not when we are at it, I like `imageheader.py` just so we keep the connection to the initial name, but don't use silly abbreviations :smile: 
* Plugin: I suppose it is still useful in itself to have the ability and showcase the options. I would move `add_image_format` directly into the plugin though, from `files.py`, as we don't use it anywhere else. Possibly just appending, now that we support all the files, and don't have exotic stuff in between?
* Testing: I mean, that would be ideal in principle, could we use `Qt` to just create as many image formats as possible with `QPixmap.save` (c.f. `tests/end2end/conftest.py`) and the existing fixtures in `conftest.py`. Fine with leaving this to an extra PR, I can try to take care of it if you like.

jcjgraf:
Glad that you like it. I also think this is a clean way to solve this problem.

I know some people who may be experienced with licensing, I will ask them if they can help us out :blush:

The question with `imageformats` is, what is its purpose? I could come up with the following answers to this question:
- It allows to add format tests for types supported by QT natively or extended (i.e. with the `qtimageformat` module installed), but which are not that popular and less frequently used (e.g. PBM, XBM, BMP, etc.). Having less tests in `tests` may make detection of actually used formats faster.
- Make a clear split between what is by supported by QT natively and by QT extended. Everything which requires the `qtimageformats` module is put into this plugin.

 At the moment it seems a bit arbitrary which formats required explicit activation via `imageformats` and which not, and which formats require the `qtimageformats` module and which not.

For reference, [these](https://doc.qt.io/qt-6/qimagereader.html#supportedImageFormats) formats are supported natively, and [these](https://doc.qt.io/qt-6/qtimageformats-index.html#the-image-i-o-plugins) are supported when having the `qtimageformats` plugin installed.

karlch:
Getting licensing experience on board would be great, thanks!

I agree, this is currently a bit arbitrary, personally I would favour this split:
- Everything supported by Qt itself, both natively and extended, should ideally be in our `imageheader` plugin directly. This would mean moving `jp2` if I see this correctly.
- User-plugins and other customs can go into `imageformats`, both `avif` and `cr2` have a user-plugin instead of something maintained by Qt if I remember correctly.

What do you think?

jcjgraf:
True; I forgot that there are also other QT modules which add support for new files types (beside `qtimageformat`).

jcjgraf:
Legal stuff is such a pain. I asked different people, and at different places and no one would give me a concrete answer (well, for obvious reasons...). The thing is that we need to maintain their license, but when we apply modifications/add own stuff, things get bit entangled and less clear. But the procedure should be (according to own research and what I was told):
- Add copyright PSF statement to file header
- Add second LICENSE file to project
- Describe in docstring of each function of that file whose work it is, and if modified, what was changed.

Anyways, since I rather write code, than "legally significant" docstrings, I decided to create a completely independent implementation of the file type detection, which I pushed to #650 . This should make it much easier for us to maintain the code in the future.

Therefore, I close this PR in favour of the other one.

PR #648, opened 2023-06-28
Improve the documentation

Yutsuten:
As I noticed on #647, updating the documentation about toggling bool setting.

Hope I'm not missing something. Again.

If the wording is bad, we can update it. I'm not english speaker, so sorry if anything sounds strange.

### DIFF
--- a/docs/documentation/configuration/keybindings.rst
+++ b/docs/documentation/configuration/keybindings.rst
@@ -7,7 +7,7 @@ to the ``$XDG_CONFIG_HOME/vimiv/keys.conf`` on first start where
 ``$XDG_CONFIG_HOME`` is usually ``~/.config/`` if you have not updated it.
 
 The configuration file is structured into sections. Each section corresponds to
-the mode in which the keybindings arevalid. In each section the keybindings are
+the mode in which the keybindings are valid. In each section the keybindings are
 defined using ``keybinding : command to bind to``. Therefore ``f : fullscreen``
 maps the ``f`` key to the ``fullscreen`` command. Special keys like space must
 be wrapped in tags in the form of ``<space>`` to allow to differentiate them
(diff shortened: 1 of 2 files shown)


Yutsuten:
By the way, somehow unrelated, but I found a typo in [docs/documentation/configuration/keybindings.rst](https://github.com/karlch/vimiv-qt/blob/master/docs/documentation/configuration/keybindings.rst?plain=1#L10C14-L10C14)

> ... the mode in which the keybindings arevalid.

If it is fine, I can fix it in this PR.

Yutsuten:
> When already at it, it would be good to also document the increment/decrement functionality for `NumberSettings` (e.g. `:set library.width +0.1`). What do you think? Could you do that too? I guess it should be something like `[[+|-]value]`.

Sure thing! I'll do that and push soon!

karlch:
Thanks a bunch, this is definitely worth documenting, good catch! Note that restructured text is notoriously annoying with indenting, way pickier than python could ever be :laughing: In your case you would need to add exactly 2 more spaces to get the formatting correct, i.e. align it with the stuff in the line above:
```
* Text.
  This goes in the same line / paragraph.
```
Your current version stops the list unfortunately. And three spaces would make the above a bold header :joy: 
```
* This is now bold.
    This goes in the next line.
```

In case you would like to check the documentation formatting before pushing
```
tox -e docs $DESTDIR
```
will build the website in `$DESTDIR`. You can then view it using `$BROWSER $DESTDIR/index.html`. :blush: 

I would also add a hint to which settings this applies, e.g. "Append a ``!`` to toggle the value of boolean settings", and "Prepend with ``+``/``-`` to increment/decrement the value of numerical settings.". What do you think?

Yutsuten:
Thank you for the guiding and suggestions. Sounds good! :+1:

Yutsuten:
Pushed the changes!

karlch:
Nice, thanks again and have a great upcoming week-end! :blush:

Issue #619, opened 2023-03-29
Unable to open WebP image format


eorus:
Is there any workaround to open webp images with vimiv? 

Thanks.

eorus:
Hi @jcjgraf ,

Thank you.. Now that I look at what you wrote, I think I've figured out the issue. However, I would like to share the details for those who will visit here later for a similar situation.

First of all, I've been using vimiv for a long time. Today I installed it directly with pacman on another machine (Archlinux). I know it's migrated to QT, but frankly, I didn't look at the version number since it was downloaded directly from community repo with several other packages. So the first problem here is that GTK with the deprecated version number "0.9.1-5" is still in the community repository.

The arch community repo should provide the "vimiv-qt 0.8.0-2" which is in active development. Otherwise, those who want to simply install vimiv like me will always go to the old version. At least for Arch :) Anyway, as a result, I installed vimiv-qt from AUR.

But after installing vimiv-qt, it still couldn't open the WebP file format I mentioned here. An additional package "qt5-imageformats" needs to be installed so that Vimiv can open WebP and perhaps other image formats. Let this be my note for the solution of the issue.

Thanks again for your quick response.

jcjgraf:
Great that you managed to get it to work.

I am sorry for the confusion. It is indeed a bit an unsatisfying situation. And you are certainly not the first person running into this issue... See e.g. #393
It is planned that the GTK version gets fully replaced once the QT ports reaches v1.0.0. Unfortunately, we have not reached that yet, as development has staggered a bit (due to time reason).

Also, it is very likely not possible to just "replace" the vimiv version in the community repo, as Arch's TUs are responsible in deciding what is included and what not.

karlch:
Thanks for your time and reporting! :blush: 

I fully agree with everything @jcjgraf has said, and unfortunately time is really the killing constraint at the moment.

In addition, I would like to mention that the AUR packages do note:
```
    'qt5-imageformats: for additional image formats'
```
do you think this message could be improved, e.g. by adding some of the formats it does support?

karlch:
Closing as we do support webp, after #650 this is pretty clean :blush:

PR #650, opened 2023-07-02
Replace imghdr

### LINKED ISSUE #579: Deprecation of imghdr module in python 3.11

jcjgraf:
replaces #646 

Since imghdr is getting deprecated from the standard library (see #579), this PR replaces that module by a own (and thereby, legally completely decoupled independent (unlike in #646)), implementation of the magic byte file type detection.

Some advantages of this approach, over using the package:
- Detect (all) file types which are actually supported by QT natively or extended (by using qtiamgeformats)
    - Well, some are not detectable...
- Detect none which are not supported
- Have a central place for all magic byte checks
    - Cleanup of `utils/files` and `plugin/imageformats`
- Well structured magic byte checks allows for better maintainability
- `plugin/imageformats` has a specific and well defined purpose
    - I.e. enable image type support that require the (manual) installation of some other QT module

And something that I like particularly well:
The runtime validation of *check functions*, (used to be done for all checks that were registered using `utils/files/add_image_formats`), is now done for all image types, except the ones natively supported by QT. This way, clients get a notification when they try to open an image of a type that:
- is supported only by qtimageformats, without having that module installed
- is supported by a 3rd party QT module and was activated via `plugins/imageformats`, without having that module installed
This should prevent confusions as in #619, and previous similar issues.

## Steps
- [x] Add check function for all types natively or extended supported
    - WBMP is completely undetectable
    - TGA version 1 is undetectable, while version 2 is
    - XBM has no signature, but can be matched based on its content
    - See docstring in `utils/imagheader` for status supported types
- [x] Add registration function
    - [x] Allow for runtime check validation
- [x] Add type detection function
- [x] Cleanup and adaption of codebase to new `utils/imageheader`
- [x] Cleanup of `plugins/imageformats`
- [x] Tests

I have already added a basic tests that verifies the detection of all filetypes which `QPixmap` is capable of writing (/creating) (as you suggested in the other PR). Unfortunately, it is capable of creating less types than what it is able to read. I have to think about ways to create and test the other types.

So, code-wise this PR is done, and if time allows, you may have a look at it. I will spend some more time for testing.

In what way does it make sense that copyright in the header of the new file is required to start at 2017? Should it not start in the year the file was first added to the project (i.e. 2023 in this case)?

fixes #579 
replaces #646

### DIFF
--- a/tests/unit/utils/test_imageheader.py
+++ b/tests/unit/utils/test_imageheader.py
@@ -1,7 +1,7 @@
 # vim: ft=python fileencoding=utf-8 sw=4 et sts=4
 
 # This file is part of vimiv.
-# Copyright 2023-2023 Christian Karl (karlch) <karlch at protonmail dot com>
+# Copyright 2017-2023 Christian Karl (karlch) <karlch at protonmail dot com>
 # License: GNU GPL v3, see the "LICENSE" and "AUTHORS" files for details.
 
 """Tests for vimiv.utils.imageheader."""
--- a/tests/unit/utils/test_imageheader.py
+++ b/tests/unit/utils/test_imageheader.py
@@ -6,22 +6,42 @@
 
 """Tests for vimiv.utils.imageheader."""
 
-from PyQt5.QtGui import QPixmap, QImageReader
+from PyQt5.QtGui import QPixmap, QImageReader, QImageWriter
 
 import pytest
 
 
 from vimiv.utils import imageheader
 
-# Image types which QPixmap is capable of creating, and not what is supported by vimiv!
-SUPPORTED_IMAGE_FORMATS = ["jpg", "png", "pbm", "pgm", "ppm", "bmp"]
+# Formats that are detectable and not detectable by this module
+DETECT_FORMATS = [format for format, _ in imageheader._registry]
+NOT_DETECT_FORMATS = ["svgz", "wbmp"]
+
+# QT Formats that alias to others
+QT_ALIAS_FORMATS = ["jpeg", "tif"]
+
+# Formats that QT can create and that are also detectable
+QT_WRITE_DETECT_FORMATS = [
+    d.data().decode()
+    for d in QImageWriter.supportedImageFormats()
+    if d.data().decode() not in NOT_DETECT_FORMATS
+    and d.data().decode() not in QT_ALIAS_FORMATS
+    and d.data().decode() != "cur"  # CUR types created by QT are actually ICO?!
+]
+
+# Formats that QT can read
+QT_READ_FORMATS = [
+    d.data().decode()
+    for d in QImageReader.supportedImageFormats()
+    if d.data().decode() not in QT_ALIAS_FORMATS
+]
 
 
 @pytest.fixture()
 def mockimageheader(mocker):
     """Fixture to mock imageheader._registry and QImageReader supportedImageFormats."""
     mocker.patch.object(
-        QImageReader, "supportedImageFormats", return_value=SUPPORTED_IMAGE_FORMATS
+        QImageReader, "supportedImageFormats", return_value=QT_READ_FORMATS
     )
     yield mocker.patch("vimiv.utils.imageheader._registry", [])
 
@@ -37,8 +57,9 @@ def create_image(filename: str, *, size=(300, 300)):
     QPixmap(*size).save(filename)
 
 
-@pytest.mark.parametrize("imagetype", SUPPORTED_IMAGE_FORMATS)
-def test_test(qtbot, tmp_path, imagetype):
+@pytest.mark.parametrize("imagetype", QT_WRITE_DETECT_FORMATS)
+def test_detect(qtbot, tmp_path, imagetype):
+    """Only tests check functions for formats that QT can create samples for."""
     name = f"img.{imagetype}"
     filename = str(tmp_path / name)
     create_image(filename)
@@ -50,7 +71,7 @@ def _check_dummy(h, f):
     return True
 
 
-@pytest.mark.parametrize("name", SUPPORTED_IMAGE_FORMATS)
+@pytest.mark.parametrize("name", QT_READ_FORMATS)
 def test_register_supported_format(mockimageheader, tmpfile, name):
     imageheader.register(name, _check_dummy)
     assert mockimageheader, "No test added by add image format"
@@ -58,6 +79,13 @@ def test_register_supported_format(mockimageheader, tmpfile, name):
 
 
 def test_register_unsupported_format(mockimageheader, tmpfile):
+    """Test if check of invalid format gets removed after call to `detect`."""
     imageheader.register("not_a_format", _check_dummy)
     assert imageheader.detect(tmpfile) is None
     assert not mockimageheader, "Unsupported test not removed"
+
+
+@pytest.mark.parametrize("name", QT_READ_FORMATS)
+def test_full_support(name):
+    """Tests if all formats readable by QT are also detected."""
+    assert name in NOT_DETECT_FORMATS or name in DETECT_FORMATS, "Missing check"
--- a/tests/unit/utils/test_imageheader.py
+++ b/tests/unit/utils/test_imageheader.py
@@ -46,13 +46,6 @@ def mockimageheader(mocker):
     yield mocker.patch("vimiv.utils.imageheader._registry", [])
 
 
-@pytest.fixture()
-def tmpfile(tmp_path):
-    path = tmp_path / "anything"
-    path.touch()
-    yield str(path)
-
-
 def create_image(filename: str, *, size=(300, 300)):
     QPixmap(*size).save(filename)
 
--- a/tests/unit/utils/test_imageheader.py
+++ b/tests/unit/utils/test_imageheader.py
@@ -1,0 +1,28 @@
+# vim: ft=python fileencoding=utf-8 sw=4 et sts=4
+
+# This file is part of vimiv.
+# Copyright 2023-2023 Christian Karl (karlch) <karlch at protonmail dot com>
+# License: GNU GPL v3, see the "LICENSE" and "AUTHORS" files for details.
+
+"""Tests for vimiv.utils.imageheader."""
+
+from PyQt5.QtGui import QPixmap
+
+import pytest
+
+
+from vimiv.utils import imageheader
+
+SUPPORTED_IMAGE_FORMATS = ["jpg", "png", "pbm", "pgm", "ppm", "bmp"]
+
+
+def create_image(filename: str, *, size=(300, 300)):
+    QPixmap(*size).save(filename)
+
+
+@pytest.mark.parametrize("imagetype", SUPPORTED_IMAGE_FORMATS)
+def test_test(qtbot, tmp_path, imagetype):
+    name = f"img.{imagetype}"
+    filename = str(tmp_path / name)
+    create_image(filename)
+    assert imageheader.detect(filename) == imagetype
--- a/tests/unit/utils/test_imageheader.py
+++ b/tests/unit/utils/test_imageheader.py
@@ -85,6 +85,14 @@ def test_register_unsupported_format(mockimageheader, tmpfile):
     assert not mockimageheader, "Unsupported test not removed"
 
 
+def test_register_unsupported_format_not_verify(mockimageheader, tmpfile):
+    """Test if check of invalid remains when registered without verification."""
+    dummy_format = "not_a_format"
+    imageheader.register(dummy_format, _check_dummy, validate=False)
+    assert imageheader.detect(tmpfile) == dummy_format
+    assert (dummy_format, _check_dummy) in imageheader._registry
+
+
 @pytest.mark.parametrize("name", QT_READ_FORMATS)
 def test_full_support(name):
     """Tests if all formats readable by QT are also detected."""
--- a/tests/unit/utils/test_imageheader.py
+++ b/tests/unit/utils/test_imageheader.py
@@ -45,7 +45,7 @@ def test_test(qtbot, tmp_path, imagetype):
     assert imageheader.detect(filename) == imagetype
 
 
-def _check_dummy(h):
+def _check_dummy(h, f):
     """Dummy image file check that always returns True."""
     return True
 
--- a/tests/unit/utils/test_imageheader.py
+++ b/tests/unit/utils/test_imageheader.py
@@ -6,14 +6,31 @@
 
 """Tests for vimiv.utils.imageheader."""
 
-from PyQt5.QtGui import QPixmap
+from PyQt5.QtGui import QPixmap, QImageReader
 
 import pytest
 
 
 from vimiv.utils import imageheader
 
+# Image types which QPixmap is capable of creating, and not what is supported by vimiv!
 SUPPORTED_IMAGE_FORMATS = ["jpg", "png", "pbm", "pgm", "ppm", "bmp"]
+
+
+@pytest.fixture()
+def mockimageheader(mocker):
+    """Fixture to mock imageheader._registry and QImageReader supportedImageFormats."""
+    mocker.patch.object(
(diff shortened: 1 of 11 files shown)


buzzingwires:
A few comments on this implementation of image detection:

* Have you considered using an existing library like [python-magic](https://github.com/ahupp/python-magic) to detect images? It is permissively licensed and libmagic has very extensive format support. libmagic should already be very commonly available on Linux systems, though I don't know how easily it will be to support that on Windows.

* For MNG files, isn't the magic number `8A 4D 4E 47 0D 0A 1A 0A`, very similar to the PNG header, but with `4D 4E 47` == `MNG` instead?

* I *think* WBMP also lack explicit headers, and I'm not sure what .cur files even are. Mouse cursor images for Windows?

* XBM and TGA indeed lack explicit magic numbers.

* For XBM files, at least the way imghdr does it is to just see if the file starts with '#define'. A better way to do it might be how libmagic does it, which is roughly equivalent to the Regex `#define [a-zA-Z0-9]+_width [0-9]+\n#define [a-zA-Z0-9]+_height [0-9]+`. Either way, XBM starts out in a pretty distinct way.

* TGA files are much less trivial, but *some* newer files *end* with `TRUEVISION-XFILE.\0` Otherwise, Targa identification seems to work based on the first 18 bytes of the file, and there are many rules to determine what generally 'makes sense' for a Targa file.

karlch:
@jcjgraf thanks a lot for the effort that went into this, this is really nice, I hope to take a closer look during this week!

Concerning the Copyright header, you are probably right, but I wouldn't bother in this PR. It may make sense to simplify this overall and remove the years and just keep a small notice at the top of each file. Discussion in a new issue?

@buzzingwires thanks a bunch for your comments and details! :blush: 

We actually did start a discussion on this in #579 . The reason for pursuing this path was initially, that just copying `imghdr` seemed absolutely straightforward and did the trick for years for us. To keep this PR on the implementation, I will add another comment in this issue directly to continue the discussion.

tests/unit/utils/test_imageheader.py
```diff
@@ -0,0 +1,63 @@
...
+    path.touch()
+    yield str(path)
+
+
+def create_image(filename: str, *, size=(300, 300)):
+    QPixmap(*size).save(filename)
+
+
+@pytest.mark.parametrize("imagetype", SUPPORTED_IMAGE_FORMATS)
+def test_test(qtbot, tmp_path, imagetype):
```
karlch, line 41: Mind calling this `test_detect` instead?

vimiv/api/__init__.py
```diff
@@ -94,5 +94,6 @@ def add_external_format(
         test_func: Function returning True if load_func supports this type.
         load_func: Function to load a QPixmap from the passed path.
     """
-    files.add_image_format(file_format, test_func)
+    # Priority to overwrite potential "default behaviour"
+    imageheader.register(file_format, test_func, priority=True)
```
karlch, line 98: The priority is needed as some CR2 used to be detected as JPG or something funky I assume? Just curious here :blush:

vimiv/utils/imageheader.py
```diff
@@ -0,0 +1,411 @@
...
+        validate: Validate that the type is supported at runtime.
+    """
+
+    @functools.wraps(check)
+    def check_verified(header: bytes) -> bool:
+        """Tests at runtime whether a filetype is actually supported.
+
+        If not, then the filetype check is unregistered.
+        """
+        if not validate:
```
karlch, line 107: I think it would be cleaner to move this check out of the actual function, i.e. register `check` directly instead of `check_verified` later, and avoid having this if in every function call.

vimiv/utils/imageheader.py
```diff
@@ -0,0 +1,411 @@
...
+    Args:
+        filename: Name of file to determine type of.
+
+    Returns:
+        Filetype or None if unknown.
+    """
+    with open(filename, "rb") as f:
+        header = f.read(32)
+
+    for filetype, check in _registry:
```
karlch, line 82: Flipping the contextlib and the for loop should in principle be possible and avoid a lot of duplicate ignored `IndexErrors`, no? I assume any file smaller than 32 is not a valid image in any case.

karlch (commented):
This is absolutely amazing work, thanks again! I only left a few nitpicks, so besides the points you mentioned and including the magic bytes described by @buzzingwires, this is basically ready to go.

It is really cool to see how much cleaner this makes the handling of filetypes, together with the Qt support, and the error / warning messages. And the documentation is amazing.

tests/unit/utils/test_imageheader.py
jcjgraf: That's not by purpose. I started writing that function when I did not know what that function is gonna be about. And once I did know, I forgot about my dummy naming 🙈

vimiv/api/__init__.py
jcjgraf: That is actually a very good point. I must have made an error in thought; The order of the check function registry has obviously no impact on the priority of QPixmap handler function in `utils.imagereader` 🙈

The second reason, which is actually still valid, is that this improves the performance of explicitly registered formats, as they are checked before all the others.

Collisions should theoretically not really happen. Compared to `imghdr` it should be even less likely, as they did sometimes some shortcuts for the checks that I did not do.

So, should I maintain this option, or change something?

karlch: It does define the QPixmap handler function though, as we pass the `file_format`, or am I having a brainfreeze?
```python
        self._handler = QImageReader(path, file_format.encode())
```

I am totally fine with keeping this, just thought we actually had some explicit reason back then as well, maybe I am wrong.

vimiv/utils/imageheader.py
jcjgraf: Makes totally sense. Simply doing:
```
    if validate:
        check = check_verified

    if priority:
        _registry.insert(0, (filetype, check))
    else:
        _registry.append((filetype, check))
```

does unfortunatelly result in some recusion issue (see testcases). Or do I miss something obvious?

Pushed a less clean solution: 7af4fbea94bcd18cc676469b20e86a04f5f32ddf

karlch: The initial option probably creates some type of recursion havoc, as the actual `check_verified` body hasn't been called yet.

How about using
```
    check_register = check_verified if validate else check
```
to make it a bit easier to parse?

For some reason this also fixes the mypy error - win-win :blush:

vimiv/utils/imageheader.py
jcjgraf: I thought about this as well. Having it as is, has certainly a negative impact on performance. I do not know if there are no valid images of size less than 32B. But given the many obscure bitmap formats, I would not be surprised if there is😅
Either we leave it as is, or change the checks such that they do not raise any exceptions, or we move the contextlib supression out of the loop and pursue one of the other two solution should ever be an issue encountered. What do you think?

vimiv/utils/imageheader.py
jcjgraf: My brain must have been to confused about the recursion weirdness that I did not think about the ternary operator 🙈 Fixed in 63f8c954b4b4600453b79c9f9cd75b15c3676cf7

vimiv/utils/imageheader.py
karlch: It seems like the only three that raise are the `pbm`, `pgm` and `ppm` on an empty file, requiring a total of 3 bytes. Now the only check we have that could require less is `bmp`, unless I read this wrong. So we could even move `bmp` ahead of `p?m` in the checklist, and then be safe moving the contextlib outside?

vimiv/api/__init__.py
jcjgraf: Actually you are right! The issue was that CR2s were detected as TIFFs (because it is based on TIFF), and then the wrong handler was used. You have such a great memory 😳🫣

While my TIFF detection is certainly better than imghdr's (uses twice the amount of information; as said, they sometimes do some simplifications), it unfortunately still matches the beginning of CR2's signature.

I have improved to comment. In addition I have prioritized the external reader in `utils.imagereader` over the base reader. This ensures that when a user registered a external handler for a format what is already supported by QT (e.g. to change some behaviour in how the image is loaded/rendered), the explicitly registered handler is used.
This is in some way similar to how the external format is prioritized over a qt format in case of a signature collision. Let me know if this makes sense to you.

See 60317ddf7752d6deb2281526f974b51b11936b89

karlch: Makes perfect sense, thanks for checking and updating! :blush:

vimiv/utils/imageheader.py
jcjgraf: You need to elaborate on this. Why cannot all other check functions raise too?

karlch: In python, slicing does not raise, you can test this with something simple like:
```python
a = []
print(a[6:12])
```

For the imageheader part, I ran this snippet:
```python
from vimiv.utils import imageheader

path = "notanimage.jpg"

with open(path, "w") as f:
    pass

with open(path, "rb") as f:
    header = f.read(32)

    for filetype, check in imageheader._registry:
        try:
            if check(header, f):
                print("filetype is", filetype)
                break
        except IndexError:
            print(f"{filetype} raises")
```
which gives:
```
pbm raises
pgm raises
ppm raises
```

jcjgraf: Hmm, I was not aware of that. But also remember that clients can register whatever check function they want 🤔

karlch: That is a valid point, maybe we should remove the suppress completely then and document that the check functions **may not raise**?

This would then re-writing the p?m checks with slicing, something along
```
    return h[:3] in [b"\x50\x31\x0A", b"\x50\x34\x0A"]
```

Now technically the jpg check could also raise if the file is only exactly the three starting bytes `FF D8 FF`, but I hope this is a technicality that doesn't exist in the wild. (I will regret saying this :joy: )

jcjgraf: Sorry for the delay...

I tried to tune the checks for performance and made them fail (i.e. return False) fast (as most of them will fail). Because of that, I tried to favour simple comparisons over more complex ones, like `in`. This may lead to less *pythonic* code, which is maybe also less readable. But more performant (hopefully; But Python behaves something weirdly when optimizing for performance).

I have done a bunch of benchmarks comparing many different implementations of the check and the detect functions. I realized that try/except induces zero overhead. And indeed, that was added in python 3.11 ([see 2. point](https://docs.python.org/3.11/whatsnew/3.11.html#misc).

Now I am wondering which approach to take:
- Create check functions that do not raise, by using slices over index access (as you suggested)
- But the checks that raise into a try/except block (I know, it is less beautiful what the contextlib supress thingy, here performance matters more that style 😃).
- Seep it as is, but replace the with contectlib stuff with a try/except block (i.e. use try for each call of the check function separately)

Let me know what you think


The benchmark I used for this:
```python
from vimiv.utils import imageheader
from typing import Optional
import timeit
import contextlib

a = "/tmp/test.bla"

with open(a, "wb") as f:
    f.write(100 * b"\x00")


def detectNative() -> Optional[str]:
    with open(a, "rb") as f:
        header = f.read(32)
        for filetype, check in imageheader._registry:
            if check(header, f):
                return filetype
    return None


def detectSupress() -> Optional[str]:
    with open(a, "rb") as f:
        header = f.read(32)
        for filetype, check in imageheader._registry:
            with contextlib.suppress(IndexError):
                if check(header, f):
                    return filetype
    return None


def detectTry() -> Optional[str]:
    with open(a, "rb") as f:
        header = f.read(32)
        for filetype, check in imageheader._registry:
            try:
                if check(header, f):
                    return filetype
            except IndexError:
                return None
    return None


if __name__ == "__main__":
    NUM_ITER = 10000
    out = timeit.timeit("detectNative()", "from __main__ import detectNative", number=NUM_ITER)
    print(out)
    out = timeit.timeit("detectSupress()", "from __main__ import detectSupress", number=NUM_ITER)
    print(out)
    out = timeit.timeit("detectTry()", "from __main__ import detectTry", number=NUM_ITER)
    print(out)

```

jcjgraf:
@buzzingwires Thanks for your really helpful comment! And sorry for the late reply...

I know that some of the formats are probably used by literally nobody (anymore). But since these formats (including `.cur`, and yes, I had also no clue that they are😅) are supported by QT, and therefore do not require any additional effort except registering a `check` function, I would like to support them.

> For MNG files, isn't the magic number 8A 4D 4E 47 0D 0A 1A 0A

Your correct, thanks! I have added a check for it.

> For XBM files, at least the way imghdr does it is to just see if the file starts with '#define'. A better way to do it might be how libmagic does it, which is roughly equivalent to the Regex #define [a-zA-Z0-9]+_width [0-9]+\n#define [a-zA-Z0-9]+_height [0-9]+. Either way, XBM starts out in a pretty distinct way.

XBM is indeed a funky file definition; it is a valid C source file🙈 I have also added a check for it.

> TGA files are much less trivial, but some newer files end with TRUEVISION-XFILE.

Indeed. There are two version of TGA. In version 2 they have added a signature to the footer of the file. I have added support for that.

> Otherwise, Targa identification seems to work based on the first 18 bytes of the file, and there are many rules to determine what generally 'makes sense' for a Targa file.

I am not really sure if it is works the effort to check for dynamic set bytes. This seems rather complex and is in the end not more that a heuristic (well, that is also the case for some other types, that have a rather "badly" chosen signature 😂), that may lead to false positives.
I think I will not add support for them here, but may do so if there is a specific request.

vimiv/utils/imageheader.py
karlch: This is interesting, I didn't know about the zero-overhead exceptions, although they do of course still induce overhead in case they are caught. However, in your last case, this is essentially like moving `contextlib.suppress` before the for loop. Once caught, we assume the file is too small to be an image, and don't catch multiple exceptions. I fully agree with this.

I am also quite surprised the `contextlib.suppress` adds so much overhead. So I would say we use the `detectTry()` version :blush: 

Thanks for the detailed investigation!!

jcjgraf: Sure there is overhead when the exception is raised. But that is not much of concern to us, as we expect that no exception is raised 😃

That is a typo. I mean to have a `pass/continue` in the except body. The reasoning is again that a check that raises should not affect other checks. This is important as clients can register whatever check function they like.

I have made these last adoptions. You may review and merge if you are happy about it 😊

karlch: Ah, yes, I always forget about the custom check functions! Thanks, looks great, I only pushed a few tiny style changes. Will merge once the tests pass :blush:

karlch (approved):

PR #652, opened 2023-07-05
Fix handling of paths with special characters / wildcards

### LINKED ISSUE #553: Bug: mark does not work if path contains "[" and "]"
It seems that an image can not be marked (with the ``mark`` command or by pressing ``m``) if the path contains both, an opening and a closing square bracket.

Some examples where marking an image doesn't work ...
```
/tmp/[a]/image.png
/tmp/[aa]/image.png
```
and where it works as usual
```
/tmp/a]/image.png
/tmp/[a/image.png
/tmp/[]/image.png
```

I've also tested the same paths with round and curly brackets. I observed no issues in these cases.

I can try to fix the issue and submit a PR if somebody can give me a hint where to start.

karlch:
Escapes the characters `[]?*` properly when expanding wildcards such as `%` and `%m`.

fixes #553 

@pglira @TeaWhyDee @jcjgraf could one of you give this a quick check?

In addition, I was not able to find out what the issue with `image.jpg~` should be, all playing around I did with tilde worked just fine. Could someone elaborate, so I can potentially add this to the PR?

### DIFF
--- a/vimiv/commands/runners.py
+++ b/vimiv/commands/runners.py
@@ -199,7 +199,7 @@ def alias(text: str, mode: api.modes.Mode) -> str:
         The replaced text if text was an alias else text.
     """
     cmd = text.split()[0]
-    if cmd in aliases.get(mode):
-        text = text.replace(cmd, aliases.get(mode)[cmd])
-        return wildcards.expand_internal(text, mode)
+    cmd_alias = aliases.get(mode).get(cmd, None)
+    if cmd_alias is not None:
+        return text.replace(cmd, cmd_alias)
     return text


jcjgraf:
I have tested the reported cases and your PR seems to fix them. I was also not able to observe anything unusual when using tiles in the file name.

TeaWhyDee:
I have tested it, the ~ is still broken. I should have specified it before,
the issue I am getting is with external commands, reproducible example command: 
```
xt : !notify-send "%%"
```
Running this on "IMG.jpg~":
image

The issue seems to persist with other special symbols too in case of external commands.

karlch:
Thanks, I can reproduce it with external commands, especially together with quoting this is complete havoc currently.

I believe we will have to clarify and change some of the requirements here.

In principle we have a matrix of combinations here, with different wildcards as the rows (internal like %, the tilde, python globbing, shell expansion) and different command types as the columns (regular command, path command such as :open, external without sub-shell using !, external with sub-shell using `:spawn`). Summarized here quick and dirty:

scrot-2023-07-06_21:35:58

yes / no indicates whether this combination **should be supported**. I am unsure on the ones colored, any feedback or discussion would be great. Here green means this is the way it is currently, red is a deviation.

Removing globbing from external commands would save us a lot of escaping-havoc, and we could strictly split shell escaping and glob escaping, as they never overlap. I am honestly not 100 % sold on changing this, yes it could be useful to run something along `!rm **/*.png`, but really this should probably be done without vimiv then. And one could still do `:mark **/*.png` followed by  `!rm %m`.

tl;dr different wildcards meet different command types, can we remove globbing from external commands? Should we add some further expansions to internal (non-path) commands?

TeaWhyDee:
I think globbing could be useful in external commands. I don't think it's currently possible to escape the globbing symbols? I think either it should be removed or \ escaping added.
Something like `:mark IMG*.jpg` and `:mark IMG\*.jpg` still marks all glob matching images (**and** this auto-completes).

As for the external commands issue, tilde and globbing expansion could probably be done before internal symbol expansion so the issue doesn't occur?

karlch:
Works with two backslashes, i.e. `:mark IMG\\*.jpg`, don't ask why :laughing: 
Now I agree it should escape in the completion, and one backslash should be enough of course.

I am not really sold on globbing in external commands though. Mainly to avoid actually using a sub-shell with `:spawn`? Or for the  additional `**` feature which not all shells have?

PR #654, opened 2023-07-06
Use a new sphinx theme for the website

karlch:
Switch from [sphinx-bootstrap-theme](https://github.com/ryan-roemer/sphinx-bootstrap-theme) to [pydata-sphinx-theme](https://github.com/pydata/pydata-sphinx-theme).

Reasons for the switch:
- Support for most recent sphinx versions (c.f. #641)
- Dark and light mode
- Clean icons in the navbar (opinionated)
- Clean navigation bar on the right (opinionated)

Reasons against the switch:
- Change of "feel"

I think the benefits outweight the cons quite clearly, but will leave this one open for a few days so anyone interested can chime in.

Quick screenshots of the new look, so one doesn't have to build it:

Startpage - light
scrot-2023-07-06_21:56:25

Screenshots - light
scrot-2023-07-06_21:56:42

Documentation - dark - showcases the navigation on the right
scrot-2023-07-06_21:57:00


jcjgraf:
I definitely prefer the new one as it looks much cleaner. Also the *On this Page* scroll bar is very usefully (actually, I only realized while comparing it to the old one, that the *Page* dropdown served that purpose on the old website). And the formatting of the commands I also like much more.
Also the search functionality is very useful (was there something like that on the old one?), though it is not perfect (yet?). It would be nice if it also showed an extract of the matching text, and not just the name of the page containing a match.

karlch:
Thanks for your feedback!

Seems like they are aware of the search limitations (c.f. https://github.com/pydata/pydata-sphinx-theme/issues/795 and https://github.com/pydata/pydata-sphinx-theme/issues/202), but progress has been limited over the past years. However, the discussion picked up steam semi-recently, so maybe we are lucky :crossed_fingers:

PR #655, opened 2023-07-07
Add the 'none' sorting setting

buzzingwires:
Starting with what I consider to be one of the simplest changes discussed in #651 and as always, nothing is as simple as expected.

It works pretty much as discussed, but it's worth noting that I found reversing the sort order does not work, probably because `lambda x: 0` doesn't provide the sorting algorithm varied numbers to meaningfully reverse. I toyed around with ways of making reversing work, but figured I'd stay with the agreed upon approach of keeping this setting simple for now.

---

For sort 'image_order' and 'directory_order', the 'none' setting can be used to allow images to follow the last order they were encountered in by the software, instead of re-sorting.

Note that setting another sorting mode while the software is running then switching back to 'none' will see files remain in the previous sort order. The original order of images is *not* tracked.

Additionally, 'none' sort cannot be reversed.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -537,3 +537,4 @@ Initial release of the Qt version.
 .. _@timsofteng: https://github.com/timsofteng
 .. _@kAldown: https://github.com/kaldown
 .. _@Yutsuten: https://github.com/Yutsuten
+.. _@buzzingwires: https://github.com/buzzingwires
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -89,6 +89,8 @@ Added:
   the changes with ``<return>`` and reject them with ``<escape>``. The
   ``{transformation-info}`` status module displays the currently selected geometry of
   the original image. Thanks `@Yutsuten`_ for reviving this!
+* Add the ``none`` sorting type for the ``sort.image_order`` and ``sort.directory_order``
+  options, implemented by `@buzzingwires`_
 
 Changed:
 ^^^^^^^^
--- a/AUTHORS
+++ b/AUTHORS
@@ -3,6 +3,7 @@ Wolfgang Popp (woefe) <mail at wolfgang-popp dot de>
 Ankur Sinha (sanjayankur31) <ankursinha at fedoraproject dot org>
 Jean-Claude Graf (jcjgraf) <jeanggi90 at gmail dot com>
 Mateus Etto (Yutsuten) <mateus dot etto at gmail dot com>
+buzzingwires (buzzingwires) <buzzingwires at outlook dot com>
 
 All contributors of non-trivial code are listed here. Please send an email to
 <karlch at protonmail dot com> or open an issue / pull request if you feel like your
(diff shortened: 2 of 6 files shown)


karlch:
Thanks, with tests, documentation and everything :heart: 

Reverse is a good catch. I do think it should support reversing, what about you? If we want to change this, we could pass a simple incrementing count as key, here is what I tested with:
```python
class Incrementer:  # Could go into utils

    def __init__(self):
        self._count = 0

    def __call__(self, _):
        self._count += 1
        return self._count


inc = Incrementer()  # Would have to be passed instead of the lambda
x = [5, 1, 3, 0, 2]

print(sorted(x))
print(sorted(x, key=inc))
print(sorted(x, key=inc, reverse=True))
```
or maybe you came up with something even simpler when toying around?

Cosmetics:
- Feel free to add yourself to the `AUTHORS` file
- Would you mind adding this to `docs/changelog.rst`, ideally thanking yourself :smile:

buzzingwires:
You're welcome!

And, reversing would be nice to have, but at least for my use case, not super-important. Another issue is that setting reverse to false doesn't set the order back to normal. It just naively passes the reversed values back through the sorting function. That's part of the reason I didn't go any further with trying to implement reversing.

The approach I took was just to call `reversed` separately, if reversing was enabled. Something like:

```python
    def sort(self, values: Iterable[str]) -> List[str]:
        """Sort values according to the current ordering."""
        ordering = self._get_ordering()
        output = sorted(values, key=ordering)
        if sort.reverse.value:
            output = list(reversed(output))
        return output
```

There might still need to be some additions to the sort function either way, and I'm not sure which would be better performance-wise, if the performance difference is significant, anyway.

And sure! Do you prefer I just add commits to the pull request or to amend and force push?

karlch:
That is a valid point, it will not properly reverse back as there is no sorting order, so I suppose having it documenting as you did is the best approach. Maybe you could add a sentence saying that the main use-case is to keep a custom input order from the commandline?

Just add some new commits :blush:

docs/changelog.rst
```diff
@@ -89,6 +89,8 @@ Added:
   the changes with ``<return>`` and reject them with ``<escape>``. The
   ``{transformation-info}`` status module displays the currently selected geometry of
   the original image. Thanks `@Yutsuten`_ for reviving this!
+* Add the ``none`` sorting type for the ``sort.image_order`` and ``sort.directory_order``
+  options, implemented by `@buzzingwires`_
```
karlch, line 93: You will also need to define the link at the bottom of the page :wink:

karlch:
Miny-nitpick in the changelog, otherwise LGTM!

Table is fine with me :blush: 
scrot-2023-07-07_14:00:40

docs/changelog.rst
buzzingwires: Oops! I'm sorry. I haven't worked much with RST files. I'll patch that right away.

karlch: Thanks! The syntax can be pretty iffy, but sphinx is just so well integrated in python that I guess it's worth it :blush:

PR #659, opened 2023-07-08
Update icons cache on manual install

Yutsuten:
Today I spent about two hours searching why when installing `vimiv-qt` manually (using `sudo make install`), vimiv's icon wouldn't appear in my application launcher (wofi).

According to [wofi man page](https://man.archlinux.org/man/wofi.7.en):

> When images are enabled drun mode will pull icon themes however being a GTK app it's possible you'll need to run gtk-update-icon-cache to get them to apply.

Then after searching about how to use `gtk-update-icon-cache` [[1]](https://manpages.ubuntu.com/manpages/xenial/man1/gtk-update-icon-cache.1.html) [[2]](https://man.archlinux.org/man/gtk4-update-icon-cache.1.en), I came up with this fix.

As I don't know if every system contains `gtk-update-icon-cache`, I allowed the command to fail by prepending it with `-`.

Hope this is an acceptable change.

### DIFF
--- a/docs/documentation/install.rst
+++ b/docs/documentation/install.rst
@@ -81,6 +81,8 @@ should work.
 
 .. include:: dependency_info.rst
 
+.. include:: updating_icon_cache.rst
+
 .. _install_using_tox:
 
 Using Tox
--- a/docs/documentation/updating_icon_cache.rst
+++ b/docs/documentation/updating_icon_cache.rst
@@ -1,0 +1,6 @@
+.. note::
+
+    You may need to run ``gtk-update-icon-cache /usr/share/icons/hicolor`` (replace the
+    path if installing the icons somewhere else) after the install to get the vimiv icon
+    in your application launcher. It has been reported that the command was needed for
+    `wofi <https://hg.sr.ht/~scoopta/wofi>`_.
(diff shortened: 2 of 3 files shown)


Yutsuten:
> However, i don't think this should be in our `Makefile`, especially as it is also used for packaging.

Sure, understandable!

I changed the documentation instead, that works too! :+1: 
Added a new "note" to the end of the "System-Wide Installation" section. Hope the wording is okay.

But now that I think about it, Linux sometimes can be very tricky. Most of my searches didn't tell about updating the icons cache, people just say "put the files here an it should be good". I only got the information about `gtk-update-icon-cache` from wofi's man page. Maybe only wofi uses this gtk caching system? Maybe no one knows about it? Or maybe its new? I don't know.

Probably when installing via AUR the problem doesn't happen because `pacman` deals with the issue for us. But when installing manually, users are probably supposed to know if they need to run something else.

Not complaining, just philosophizing that in the Linux world we need to learn something new every day. :grin:

karlch:
This looks good to me, happy to merge!

So arch definitely knows about it, there is a pacman hook specifically for this purpose, see `/usr/share/libalpm/hooks/gtk-update-icon-cache.hook`. This is actually precisely the reason why I wouldn't put this in the `Makefile`, the system may/should take care of it.

Yutsuten:
> happy to merge unless you have further questions / topics to be addressed

I'm done with the PR, you can merge it any time. Thank you!

Issue #335, opened 2021-01-31
Randomly failing / cancelled CI runners


karlch:
From time to time, more often than not at times, one or more runners in the CI are extremely slow and get killed by the timeout after running for ages. This is extremely annoying as the state of the CI then is failed, although there is nothing wrong with the code.

I have only seen this happen with the pytest runners, which means there may be something we can do on our side.

PR #660, opened 2023-07-11
Add thumbnail.save configuration option.

buzzingwires:
Continuing #651 changes. Also, the `check_thumbails_created` in `tests/unit/utils/test_thumbnail_manager.py` might be a typo. Figured this should probably be another pull request or quick commit from maintainers, though.

---

If this option is set to False, then do not save newly-generated thumbnails to the disk. Previously saved thumbnail images will be used normally and will never be deleted.

This saves space and reduces drive writes at the cost of extra computational overhead by making the original image be reopened each time the thumbnail is loaded.

### DIFF
--- a/tests/integration/test_read_settings.py
+++ b/tests/integration/test_read_settings.py
@@ -19,7 +19,7 @@ from vimiv.config import configfile
 UPDATED_CONFIG = {
     "SORT": {"shuffle": "True"},
     "IMAGE": {"overzoom": "4.2"},
-    "THUMBNAIL": {"size": "64", "save": "False"},
+    "THUMBNAIL": {"size": "64"},
 }
 
 
@@ -46,7 +46,7 @@ UPDATED_EXIF_KEY_SETS = {
 UPDATED_CONFIG_INVALID = {
     "SORT": {"shuffle": "not a bool"},
     "IMAGE": {"overzoom": "not a float"},
-    "THUMBNAIL": {"size": "not an int", "save": "not a bool"},
+    "THUMBNAIL": {"size": "not an int"},
 }
 
 
@@ -86,7 +86,6 @@ def test_read_config(configpath):
     assert api.settings.sort.shuffle.value is True
     assert api.settings.image.overzoom.value == 4.2
     assert api.settings.thumbnail.size.value == 64
-    assert api.settings.thumbnail.save.value is False
 
 
 @pytest.mark.parametrize(
--- a/tests/integration/test_read_settings.py
+++ b/tests/integration/test_read_settings.py
@@ -19,7 +19,7 @@ from vimiv.config import configfile
 UPDATED_CONFIG = {
     "SORT": {"shuffle": "True"},
     "IMAGE": {"overzoom": "4.2"},
-    "THUMBNAIL": {"size": "64"},
+    "THUMBNAIL": {"size": "64", "save": "False"},
 }
 
 
@@ -46,7 +46,7 @@ UPDATED_EXIF_KEY_SETS = {
 UPDATED_CONFIG_INVALID = {
     "SORT": {"shuffle": "not a bool"},
     "IMAGE": {"overzoom": "not a float"},
-    "THUMBNAIL": {"size": "not an int"},
+    "THUMBNAIL": {"size": "not an int", "save": "not a bool"},
 }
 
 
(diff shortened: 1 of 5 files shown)


tests/integration/test_read_settings.py
```diff
@@ -46,7 +46,7 @@
 UPDATED_CONFIG_INVALID = {
     "SORT": {"shuffle": "not a bool"},
     "IMAGE": {"overzoom": "not a float"},
-    "THUMBNAIL": {"size": "not an int"},
+    "THUMBNAIL": {"size": "not an int", "save": "not a bool"},
```
karlch, line 49: Same as above.

tests/integration/test_read_settings.py
```diff
@@ -86,6 +86,7 @@ def test_read_config(configpath):
     assert api.settings.sort.shuffle.value is True
     assert api.settings.image.overzoom.value == 4.2
     assert api.settings.thumbnail.size.value == 64
+    assert api.settings.thumbnail.save.value is False
```
karlch, line 89: Same as above.

tests/unit/utils/test_thumbnail_manager.py
```diff
@@ -24,6 +25,28 @@ def manager(qtbot, tmp_path, mocker):
...
+    has_thumbnail_path = str(tmp_path / "has_thumbnail.jpg")
+    QPixmap(300, 300).save(has_thumbnail_path, "jpg")
+    manager.create_thumbnails_async([has_thumbnail_path])
+    check_thumbails_created(qtbot, manager, 1)
+
+    monkeypatch.setattr(settings.thumbnail.save, "value", False)
+    no_thumbnail_path = str(tmp_path / "no_thumbnail.jpg")
+    QPixmap(300, 300).save(no_thumbnail_path, "jpg")
+    manager.create_thumbnails_async([has_thumbnail_path, no_thumbnail_path])
+    check_thumbails_created(qtbot, manager, 1)
```
karlch, line 47: Could you elaborate why you believe there is a typo involved?

The function does not start with `test_` as it is not a standalone unit-test, but instead used by various tests to perform the same assert + wait logic.

vimiv/api/settings.py
```diff
@@ -470,6 +470,9 @@ class thumbnail:  # pylint: disable=invalid-name
     """Namespace for thumbnail related settings."""
 
     size = ThumbnailSizeSetting("thumbnail.size", 128, desc="Size of thumbnails")
+    save = BoolSetting(
+        "thumbnail.save", True, desc="Save new thumbnails to disk for later use"
```
karlch, line 474: Maybe include the shared icon cache somehow?

Save new thumbnails to disk in the shared icon cache?

karlch (changes requested):
Thanks a bunch, once again with tests and everything :blush: 

I added a few tiny nitpicks + questions, should be trivial to solve.

tests/unit/utils/test_thumbnail_manager.py
buzzingwires: Oh, if you see the function `check_thumbails_created`, thumbails is missing the n

tests/integration/test_read_settings.py
```diff
@@ -19,7 +19,7 @@
 UPDATED_CONFIG = {
     "SORT": {"shuffle": "True"},
     "IMAGE": {"overzoom": "4.2"},
-    "THUMBNAIL": {"size": "64"},
+    "THUMBNAIL": {"size": "64", "save": "False"},
```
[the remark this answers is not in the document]
buzzingwires: I'm sorry, I now understand the purpose of these tests. Removing the extra tests isn't a problem.

tests/unit/utils/test_thumbnail_manager.py
karlch: Ah, damnit, must have been blind twice, thanks :laughing:

vimiv/api/settings.py
buzzingwires: Rephrased this description. Please let me know if it needs any further changes.

buzzingwires:
Looks like one of the tests timed out at `tests/end2end/features/edit/test_transform_bdd.py` ?

vimiv/api/settings.py
karlch: Thanks! Not sure about "to the disk" compared to "to disk", but can also just change this after merge along with the typo you pointed out.

karlch (approved):

karlch:
Yes, unfortunately the runner is extremely slow from time to time (one can see that actually all tests progress a lot slower). I have absolutely no clue why, this only happens on CI, haven't had any issues locally. Noted a while ago in #335.

PR #665, opened 2023-07-14
Implement lazy loading of thumbnails

buzzingwires:
Now for one of the more ambitious changes discussed in #651 The general goal is to allow thumbnails to be displayed in a lazy-loaded manner, and unloaded when they fall out of a certain range. The number of thumbnails displayed at any given time being configurable to this end by the following options:

* `thumbnail.load_behind` : Max thumbnails to render behind the selected one. -1 (default) is no limit
* `thumbnail.load_ahead`  : Max thumbnails to render ahead of the selected one. -1 (default) is no limit
* `thumbnail.unload_threshold` : Don't unload thumbnails unless more are loaded than this. -1 (default) is no limit

### DIFF
--- a/tests/end2end/features/thumbnail/test_thumbnaillazy.feature
+++ b/tests/end2end/features/thumbnail/test_thumbnaillazy.feature
@@ -3,10 +3,11 @@ Feature: Lazy load thumbnails
     Scenario: Eager loading (default behavior)
         Given I open 10 images without leading zeros in their name
         And I enter thumbnail mode
+        Then there should be 10 rendered thumbnails
         When I run 5scroll right
-        Then the first index should be 0
+        Then there should be 10 rendered thumbnails
+        And the first index should be 0
         And the last index should be 9
-        And there should be 10 rendered thumbnails
 
     Scenario: Only display the selected thumbnail
         Given I open 10 images without leading zeros in their name
--- a/tests/end2end/features/thumbnail/test_thumbnaillazy.feature
+++ b/tests/end2end/features/thumbnail/test_thumbnaillazy.feature
@@ -62,4 +62,4 @@ Feature: Lazy load thumbnails
         Then there should be 6 rendered thumbnails
         And the first index should be 3
         And the last index should be 7
-		And image_8.jpg should be in the rendered thumbnails
+        And image_8.jpg should be in the rendered thumbnails
--- a/tests/end2end/features/thumbnail/test_thumbnaillazy.feature
+++ b/tests/end2end/features/thumbnail/test_thumbnaillazy.feature
@@ -1,0 +1,65 @@
+Feature: Lazy load thumbnails
+
+    Scenario: Eager loading (default behavior)
+        Given I open 10 images without leading zeros in their name
+        And I enter thumbnail mode
+        When I run 5scroll right
+        Then the first index should be 0
+        And the last index should be 9
+        And there should be 10 rendered thumbnails
+
+    Scenario: Only display the selected thumbnail
+        Given I open 10 images without leading zeros in their name
+        And I enter thumbnail mode
+        Then there should be 10 rendered thumbnails
+        When I run set thumbnail.load_behind 0
+        When I run set thumbnail.load_ahead 0
+        When I run set thumbnail.unload_threshold 0
+        When I run 5scroll right
+        Then there should be 1 rendered thumbnails
+        And the first index should be 5
+        And the last index should be 5
+
+    Scenario: Display two thumbnails behind and all ahead
+        Given I open 10 images without leading zeros in their name
+        And I enter thumbnail mode
+        Then there should be 10 rendered thumbnails
+        When I run set thumbnail.load_behind 2
+        When I run 5scroll right
+        Then there should be 7 rendered thumbnails
+        And the first index should be 3
+        And the last index should be 9
+
+    Scenario: Display two thumbnails ahead and all behind
+        Given I open 10 images without leading zeros in their name
+        And I enter thumbnail mode
+        Then there should be 10 rendered thumbnails
+        When I run set thumbnail.load_ahead 2
+        When I run 5scroll right
+        Then there should be 8 rendered thumbnails
+        And the first index should be 0
+        And the last index should be 7
+
+    Scenario: Display two thumbnails ahead and behind
+        Given I open 10 images without leading zeros in their name
+        And I enter thumbnail mode
+        Then there should be 10 rendered thumbnails
+        When I run set thumbnail.load_ahead 2
+        When I run set thumbnail.load_behind 2
+        When I run 5scroll right
+        Then there should be 5 rendered thumbnails
+        And the first index should be 3
+        And the last index should be 7
+
+    Scenario: Display two thumbnails ahead and behind but don't unload less than six tumbnails
+        Given I open 10 images without leading zeros in their name
+        And I enter thumbnail mode
+        Then there should be 10 rendered thumbnails
+        When I run set thumbnail.load_ahead 2
+        When I run set thumbnail.load_behind 2
+        When I run set thumbnail.unload_threshold 6
+        When I run 5scroll right
+        Then there should be 6 rendered thumbnails
+        And the first index should be 3
+        And the last index should be 7
+		And image_8.jpg should be in the rendered thumbnails
--- a/vimiv/api/settings.py
+++ b/vimiv/api/settings.py
@@ -475,6 +475,8 @@ class thumbnail:  # pylint: disable=invalid-name
         True,
         desc="Save new thumbnails to disk in the shared icon cache for later use",
     )
+    max_behind = IntSetting("thumbnail.max_behind", 50, desc="Maximum number of thumbnails to render behind the currently selected one.")
+    max_ahead = IntSetting("thumbnail.max_ahead", 50, desc="Maximum number of thumbnails to render ahead of the currently selected one.")
 
 
 class slideshow:  # pylint: disable=invalid-name
--- a/vimiv/api/settings.py
+++ b/vimiv/api/settings.py
@@ -475,9 +475,22 @@ class thumbnail:  # pylint: disable=invalid-name
         True,
         desc="Save new thumbnails to disk in the shared icon cache for later use",
     )
-    load_behind = IntSetting("thumbnail.load_behind", -1, desc="Maximum number of thumbnails to render behind the currently selected one. -1 means no limit.")
-    load_ahead = IntSetting("thumbnail.load_ahead", -1, desc="Maximum number of thumbnails to render ahead of the currently selected one. -1 means no limit.")
-    unload_threshold = IntSetting("thumbnail.unload_threshold", -1, desc="Don't unload thumbnails unless there are more than this number rendered. -1 means no limit.")
+    load_behind = IntSetting(
+        "thumbnail.load_behind",
+        -1,
+        desc="Max thumbnails to render behind the selected one. -1 is no limit",
+    )
+    load_ahead = IntSetting(
+        "thumbnail.load_ahead",
+        -1,
+        desc="Max thumbnails to render ahead of the selected one. -1 is no limit",
+    )
+    unload_threshold = IntSetting(
+        "thumbnail.unload_threshold",
+        -1,
+        desc="Don't unload thumbnails unless more are loaded than this. -1 is no limit",
+    )
+
 
 class slideshow:  # pylint: disable=invalid-name
     """Namespace for slideshow related settings."""
--- a/vimiv/api/settings.py
+++ b/vimiv/api/settings.py
@@ -475,9 +475,9 @@ class thumbnail:  # pylint: disable=invalid-name
         True,
         desc="Save new thumbnails to disk in the shared icon cache for later use",
     )
-    max_behind = IntSetting("thumbnail.max_behind", 50, desc="Maximum number of thumbnails to render behind the currently selected one.")
-    max_ahead = IntSetting("thumbnail.max_ahead", 50, desc="Maximum number of thumbnails to render ahead of the currently selected one.")
-    max_count = IntSetting("thumbnail.max_count", 200, desc="Maximum number of thumbnails to render in general.")
+    max_behind = IntSetting("thumbnail.max_behind", 0, desc="Maximum number of thumbnails to render behind the currently selected one.")
+    max_ahead = IntSetting("thumbnail.max_ahead", 0, desc="Maximum number of thumbnails to render ahead of the currently selected one.")
+    max_count = IntSetting("thumbnail.max_count", 0, desc="Maximum number of thumbnails to render in general.")
 
 class slideshow:  # pylint: disable=invalid-name
     """Namespace for slideshow related settings."""
--- a/vimiv/api/settings.py
+++ b/vimiv/api/settings.py
@@ -475,9 +475,9 @@ class thumbnail:  # pylint: disable=invalid-name
         True,
         desc="Save new thumbnails to disk in the shared icon cache for later use",
     )
-    load_behind = IntSetting("thumbnail.load_behind", 0, desc="Maximum number of thumbnails to render behind the currently selected one.")
-    load_ahead = IntSetting("thumbnail.load_ahead", 0, desc="Maximum number of thumbnails to render ahead of the currently selected one.")
-    unload_threshold = IntSetting("thumbnail.unload_threshold", 0, desc="Don't unload thumbnails unless there are more than this number rendered.")
+    load_behind = IntSetting("thumbnail.load_behind", -1, desc="Maximum number of thumbnails to render behind the currently selected one. -1 means no limit.")
+    load_ahead = IntSetting("thumbnail.load_ahead", -1, desc="Maximum number of thumbnails to render ahead of the currently selected one. -1 means no limit.")
+    unload_threshold = IntSetting("thumbnail.unload_threshold", -1, desc="Don't unload thumbnails unless there are more than this number rendered. -1 means no limit.")
 
 class slideshow:  # pylint: disable=invalid-name
     """Namespace for slideshow related settings."""
--- a/vimiv/api/settings.py
+++ b/vimiv/api/settings.py
@@ -477,7 +477,7 @@ class thumbnail:  # pylint: disable=invalid-name
     )
     max_behind = IntSetting("thumbnail.max_behind", 50, desc="Maximum number of thumbnails to render behind the currently selected one.")
     max_ahead = IntSetting("thumbnail.max_ahead", 50, desc="Maximum number of thumbnails to render ahead of the currently selected one.")
-
+    max_count = IntSetting("thumbnail.max_count", 200, desc="Maximum number of thumbnails to render in general.")
 
 class slideshow:  # pylint: disable=invalid-name
     """Namespace for slideshow related settings."""
--- a/vimiv/api/settings.py
+++ b/vimiv/api/settings.py
@@ -475,9 +475,9 @@ class thumbnail:  # pylint: disable=invalid-name
         True,
         desc="Save new thumbnails to disk in the shared icon cache for later use",
     )
-    max_behind = IntSetting("thumbnail.max_behind", 0, desc="Maximum number of thumbnails to render behind the currently selected one.")
-    max_ahead = IntSetting("thumbnail.max_ahead", 0, desc="Maximum number of thumbnails to render ahead of the currently selected one.")
-    max_count = IntSetting("thumbnail.max_count", 0, desc="Maximum number of thumbnails to render in general.")
+    load_behind = IntSetting("thumbnail.load_behind", 0, desc="Maximum number of thumbnails to render behind the currently selected one.")
+    load_ahead = IntSetting("thumbnail.load_ahead", 0, desc="Maximum number of thumbnails to render ahead of the currently selected one.")
+    unload_threshold = IntSetting("thumbnail.unload_threshold", 0, desc="Don't unload thumbnails unless there are more than this number rendered.")
 
 class slideshow:  # pylint: disable=invalid-name
     """Namespace for slideshow related settings."""
--- a/vimiv/gui/thumbnail.py
+++ b/vimiv/gui/thumbnail.py
@@ -83,6 +83,7 @@ class ThumbnailView(
         widgets.ScrollWheelCumulativeMixin.__init__(self, self._scroll_wheel_callback)
         QListWidget.__init__(self)
 
+        self._rendered_paths = set()
         self._paths: List[str] = []
 
         fail_pixmap = create_pixmap(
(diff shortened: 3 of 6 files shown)


tests/end2end/features/thumbnail/test_thumbnaillazy.feature
```diff
@@ -0,0 +1,66 @@
...
+        Then there should be 10 rendered thumbnails
+        When I run 5scroll right
+        Then there should be 10 rendered thumbnails
+        And the first index should be 0
+        And the last index should be 9
+
+    Scenario: Only display the selected thumbnail
+        Given I open 10 images without leading zeros in their name
+        And I enter thumbnail mode
+        Then there should be 10 rendered thumbnails
```
karlch, line 15: I would remove this `Then`, as it is essentially tested by the previous test, no? Same for the following tests.

tests/end2end/features/thumbnail/test_thumbnaillazy.feature
```diff
@@ -0,0 +1,66 @@
...
+        When I run set thumbnail.load_behind 0
+        When I run set thumbnail.load_ahead 0
```
karlch, line 17: You could consider replacing multiple subsequent `When` statements with `When`, `And`, `And`, ... This is more consistent with the rest of the tests, but not a deal-breaker if you prefer it as-is.

tests/end2end/features/thumbnail/test_thumbnaillazy.feature
```diff
@@ -0,0 +1,66 @@
+Feature: Lazy load thumbnails
```
karlch, line 1: Please rename the file to `thumbnaillazy.feature` instead of prepending `test_` :blush:

vimiv/api/settings.py
```diff
@@ -475,6 +475,21 @@ class thumbnail:  # pylint: disable=invalid-name
...
+    )
+    load_ahead = IntSetting(
+        "thumbnail.load_ahead",
+        -1,
+        desc="Max thumbnails to render ahead of the selected one. -1 is no limit",
+    )
+    unload_threshold = IntSetting(
+        "thumbnail.unload_threshold",
+        -1,
+        desc="Don't unload thumbnails unless more are loaded than this. -1 is no limit",
```
karlch, line 491: Maybe we can somehow specify that they are of course unloaded when the filelist changes? Something along
"Don't unload thumbnails in the current selection unless ... ". Selection is probably not the best word, maybe you have a better one in mind :blush:

vimiv/gui/thumbnail.py
```diff
@@ -143,6 +144,75 @@ def n_rows(self) -> int:
         """Return the number of rows."""
         return math.ceil(self.count() / self.n_columns())
 
+    def _first_index(self) -> int:
```
karlch, line 147: Nitpick, but consider renaming to `_first_rendered_index` for clarification. Same for `_last` later.

vimiv/gui/thumbnail.py
```diff
@@ -143,6 +144,75 @@ def n_rows(self) -> int:
...
+        """Return the index of the first thumbnail to be rendered."""
+        result: int
+        if api.settings.thumbnail.load_behind.value == -1:
+            result = 0
+        else:
+            result = max(
+                0, self.current_index() - api.settings.thumbnail.load_behind.value
+            )
+        return result
```
karlch, line 156: Personal preference and consistency:
```python
if ...
    return 0
return max(...)

 Same for `_last` later.

File: vimiv/gui/thumbnail.py [hunk 7]
```diff
@@ -143,6 +144,75 @@ def n_rows(self) -> int:
...
+        elif api.settings.thumbnail.load_ahead.value == -1:
+            result = len(self._paths) - 1
+        else:
+            result = min(
+                len(self._paths) - 1,
+                self.current_index() + api.settings.thumbnail.load_ahead.value,
+            )
+        return result
+
+    def _index_range(self) -> Tuple[int, int]:
```
karlch, line 172: `_rendered_index_range`?

vimiv/gui/thumbnail.py
```diff
@@ -143,6 +144,75 @@ def n_rows(self) -> int:
...
+            self._rendered_paths.discard(self._paths[index])
+
+    def _prune_icons_behind(self) -> None:
+        """Prune indexes before the current index."""
+        for index in range(0, self._first_index()):
+            self._prune_index(index)
+
+    def _prune_icons_ahead(self) -> None:
+        """Prune indexes after the current index."""
+        for index in reversed(range(self._last_index() + 1, len(self._paths))):
```
karlch, line 194: Out of curiosity, why the `reversed`?

vimiv/gui/thumbnail.py
```diff
@@ -238,21 +313,25 @@ def open_selected(self):
         api.signals.load_images.emit([self.current()])
         api.modes.IMAGE.enter()
 
-    @api.keybindings.register("<ctrl>b", "scroll page-up", mode=api.modes.THUMBNAIL)
-    @api.keybindings.register("<ctrl>f", "scroll page-down", mode=api.modes.THUMBNAIL)
+    @api.keybindings.register(
+        ("<ctrl>b", "<page-up>"), "scroll page-up", mode=api.modes.THUMBNAIL
```
karlch, line 317: While I don't see any harm in these keybindings, they should go into a new PR in case you want them merged upstream :blush: If merged upstream, they should probably also be included in the library for consistency.

vimiv/utils/thumbnail_manager.py
```diff
@@ -72,14 +72,15 @@ def __init__(self, fail_pixmap: QPixmap, large: bool = True):
         xdg.makedirs(self.directory, self.fail_directory)
         self.fail_pixmap = fail_pixmap
 
-    def create_thumbnails_async(self, paths: List[str]) -> None:
+    def create_thumbnails_async(self, indices: Iterable[int], paths: List[str]) -> None:
```
karlch, line 75: In the current implementation the indices are not always subsequent due to the threshold, right? Otherwise I would have said `paths: List[str], start: int = 0` and a later `enumerate` would be a lot easier, and make our life in the tests easier as well (passing the `[0]`, ... in all unit tests seems a bit clumsy.)

This does make me think if the `threshold` is really that useful, do you have it set above the interval size?

karlch (changes requested):
Thanks a lot, this is excellent work, with really nice test-cases!

I did leave a few nit-picking comments, the one part that once again requires a bit of thinking is the `threshold` as it does seem to make the implementation a tad more clumsy. Would love to hear your thoughts :blush: 

Have a great week-end!

karlch:
Besides the code comments, I have noticed some havoc, e.g. when scrolling fast (holding `j` or `k`) on a larger list of thumbnails, resulting in something like this (`unload_threshold=0`, `load_*=5`, so I would expect 5 ahead and 5 behind, nothing else):

scrot-2023-07-15_11:44:43

Could you investigate?

vimiv/gui/thumbnail.py
buzzingwires: The `reversed` is there, so that when pruning icons ahead of the current selection, it's done from the end first. If it went in normal order, there would be a gap from the icons removed after the selection. I hope I explained that reasonably well.

vimiv/gui/thumbnail.py
buzzingwires: The reason I made those keybindings is that scrolling with `page-up`, `page-down`, arrow keys, and so on would not trigger the scroll function, and thus the icons would not be updated to the new position. I wonder if this scrolling behavior is a byproduct of the OS' standard window controls. Could maybe be overridden somewhere, though?

tests/end2end/features/thumbnail/test_thumbnaillazy.feature
buzzingwires: The use of the `Then` there is admittedly a bit of a hack. You'll note that in the implementation of these narrative steps, there's the `wait_for_thumbnails_to_load` called at the beginning of every step. The tests don't behave unless the thumbnail rendering is finished before changing the settings.

It appears there's some kind of race condition, and I wonder if a similar race condition contributes to the bug you mention below.

vimiv/utils/thumbnail_manager.py
buzzingwires: Yes, I think I might have actually tried something like you described, before settling on this. And right, the setting should really only make a difference when set above the interval size.

It's intended for say... suddenly jumping to a position in the file list, then jumping back to the previous location, or something like that. Since there can be images left loaded out of range, they aren't suddenly loaded again in short order.

Granted, there would still be the delay of loading images from that sudden jump, in the first place, just not the jump back. And reads aren't a big deal on an SSD.

buzzingwires:
You're welcome! I'm glad you generally like how this is turning out. You have a good weekend, too!

I most-certainly will investigate the bug. I am not *immediately* able to reproduce it, but as mentioned in the review for one of the tests, I think there's some kind of race condition that could be causing this.

vimiv/api/settings.py
buzzingwires: I'm going for something like... "Don't unload currently rendered thumbnails unless more are loaded than this. -1 is no limit" ?

karlch: Nice, maybe even "unless above this threshold"?

buzzingwires: I leaned to something slightly more verbose: `Don't unload currently rendered thumbnails unless **the number loaded is** above this threshold.` Can do the other, though. Or just do 'the number is', since it will refer back to 'currently rendered thumbnails'.

buzzingwires:
Think I fixed our race condition and understand why it happened. Fixing this also removes the need for the redundant check for rendered thumbnails in the tests. See e44a9383547eb1c4689c9a8794b45d814d914b47

tests/end2end/features/thumbnail/test_thumbnaillazy.feature
buzzingwires: Fixed by e44a9383547eb1c4689c9a8794b45d814d914b47

buzzingwires:
Hmmmn, now we have one of the tests passing on my machine, but failing on the remote.

karlch:
Quick thanks for all the work you are putting in and investigations you are doing! As a heads-up: I will unfortunately not be able to take a deeper look until the week-end, as I am fully busy with a business trip during the week.

buzzingwires:
Leaving my notes here:

The number of images to be loaded being less than the number of threads set for the pool is what seems to cause this to happen. I believe QT basically works by issuing a job to each thread and waiting for those jobs to finish before assigning new ones.

When navigating the thumbnails, async thumbnail load is called with a new set of thumbnails when the position changes. At this time, the current behavior is to clear the pool to remove queued threads. So, when the number of threads is less than the number of jobs, this means there are threads in the queue that will never be started, and thus, they will never be loaded and checked against the threshold.

I'm able to get deterministic behavior by not clearing the pool, but that will create a greater performance hit from always loading every queued thread. So, we have a variety of choices here, and I'm inclined to wonder whether this deterministic behavior is truly necessary.

vimiv/utils/thumbnail_manager.py
buzzingwires: The way indices are specified in unit tests is also cleaned up a little by 8d9118e358c0a9d757817e4d77ade834fa16c1d9

vimiv/api/settings.py
karlch: Sounds perfect to me, thanks!

vimiv/gui/thumbnail.py
karlch: I see, those are most likely default bindings of the `QListWidget` then. If we don't recognize a binding, we try to pass it to Qt. Makes sense to override in this case :+1:

vimiv/utils/thumbnail_manager.py
karlch: Perfect, thank you!

tests/end2end/features/thumbnail/thumbnaillazy.feature
```diff
@@ -0,0 +1,66 @@
+Feature: Lazy load thumbnails
+
+    Scenario: Eager loading (default behavior)
+        Given I open 10 images without leading zeros in their name
+        And I enter thumbnail mode
+        When I run 5scroll right
+        Then there should be 10 rendered thumbnails
+        And the first index should be 0
+        And the last index should be 9
+		And thumbnails should load in order of closeness to selected
```
karlch, line 10: Tabs instead of spaces I guess? Just leaving this here before I forget, will continue understanding the behaviour and leave an overall comment.

vimiv/gui/thumbnail.py
```diff
@@ -143,12 +144,86 @@ def n_rows(self) -> int:
...
+        return (self._first_rendered_index(), self._last_rendered_index() + 1)
+
+    def _rendered_thumbnail_pairs(self) -> List[Tuple[int, str]]:
+        """Return index/path pairs with those closest to current index first."""
+        current_index = self.current_index()
+        first_index, last_index = self._rendered_index_range()
+        pairs = list(
+            zip(range(first_index, last_index), self._paths[first_index:last_index])
+        )
+        pairs.sort(key=lambda p: abs(p[0] - current_index))
```
karlch, line 175: Another random comment, should have probably started another review, sorry, there may be more coming :laughing: 

The combination of passing this as a list of tuples, which is later always unpacked and zipped seems .. over-the-top. I would definitely pass this as a dictionary instead, so we can later use `list(dict)`, `dict.items()`, and `dict.values()` instead of dealing with the list by hand.

First idea would be something along:
```python
indices = sorted(range(first_index, last_index), key=lambda i: abs(i - current_index))
return dict(zip(indices, self._paths[indices]))
```

vimiv/gui/thumbnail.py
```diff
@@ -446,6 +532,7 @@ def resizeEvent(self, event):
         """Update resize event to keep selected thumbnail centered."""
         super().resizeEvent(event)
         self.scrollTo(self.currentIndex())
+        self._update_icon_position()
```
karlch, line 535: Could you elaborate why this is needed? Upon resizing the widget / window, we do have to update scrolling, as the view may have changed, but we shouldn't have to generate the thumbnails again, or am I missing something?

Noticed that this was called a lot more in testing than I thought it would due to this part.

karlch:
So, I believe I finally understood what is happening, thanks for your work and explanations! I agree that it should not be deterministic which part is affected by the `unload_threshold`.

I however do think we should ensure that the two ahead and two behind are rendered. And this is perfectly the case.

I would rewrite the tests along:
```bdd
Feature: Lazy load thumbnails

    Background:
        Given I open 10 images
        And I enter thumbnail mode

    Scenario: Eager loading (default behavior)
        When I run 5scroll right
        Then there should be 10 rendered thumbnails
        And the first index should be 0
        And the last index should be 9
        And thumbnails should load in order of closeness to selected
...
    Scenario: Display two thumbnails ahead and behind but don't unload less than six tumbnails
        When I run set thumbnail.load_ahead 2
        And I run set thumbnail.load_behind 2
        And I run set thumbnail.unload_threshold 6
        And I run 5scroll right
        Then there should be 6 rendered thumbnails
        And the first index should be 3
        And the last index should be 7
        And image_08.jpg should be in the rendered thumbnails
        And thumbnails should load in order of closeness to selected
```

I am not sure why you decided to explicitly avoid the trailing zeros, as it makes the ordering quite confusing, at least for me. I.e., `image_1.jpg`, `image_10.jpg`, `image_2.jpg`, and so forth.

As I am still unsure I understand the pruning and so forth correctly everywhere, I may need another deep-dive into the code, but really feel like we are converging towards the expected behaviour :blush:

vimiv/gui/thumbnail.py
buzzingwires: I honestly do not remember why I added this. It might had just been that I overzealously added it to every function I felt could have influenced the selected thumbnail. At least during testing, removing this doesn't seem to make a difference, though. So, I will do that.

vimiv/gui/thumbnail.py
buzzingwires: Your approach is definitely better. More concise, and probably more performant as well, without the extra unpacking. I just needed to replace `self._paths[indices]` with a list comprehension to select the appropriate paths.

buzzingwires:
Rewrote the tests as specified and ended up removing the test for a thumbnail at a specific index, because that's really the only part that's non-deterministic.

There wasn't really a great reason for stripping the trailing zeroes, to be honest. Somewhere in my brain at the time, I may had conceived it would be more concise to remove them, and I was trying to match another test case I was using as reference. Anyway, they're not stripped now.

Also added a little bit of code that allows the icon position update to happen if moving around by clicking, rather than by scrolling or using the keyboard.

And hey, thanks! This has been a slog, so I appreciate your being patient with me. Is there anything I could help clarify concerning the pruning algorithm?

vimiv/gui/thumbnail.py
karlch: You are right of course, this is not a numpy array :smile: 

Although, we could then go straight for a dict-comprehension, but this is becoming super nit-picky:
```
        return {i: self._paths[i] for i in indices}
```

karlch:
Perfect, thank you so much!

The one thing I am currently not able to wrap my head around, is if we need to call the pruning once a single new thumbnail has been created. Probably, yes, but it does seem to cause some overhead, even when nothing is pruned. Couldn't pruning be done once scrolling? Sure, there may be short times with "too little" thumbnails, but those will be filled then again.

vimiv/gui/thumbnail.py
buzzingwires: I honestly didn't even know dict comprehensions were a thing. So, thank you for teaching me something new. Have another commit ready to push with the change, now.

buzzingwires:
Of course! I'm glad you're ultimately happy with the way this is turning out.

If I recall correctly, I originally tried to do the pruning after scrolling, but there was a race condition in which the scroll function would prune thumbnails and then return before the new thumbnail was actually loaded. Since the new thumbnail wasn't loaded yet, old ones that should have been pruned sometimes weren't.

karlch:
That's unfortunate. I also suppose it doesn't make sense to try to prune once the current set of thumbnails has been created? There doesn't seem to be something along `QThreadPool.done` to connect to, but we could possibly hack something ourselves, if and only if this implementation would work at all :smile:

buzzingwires:
https://github.com/karlch/vimiv-qt/commit/15610c3be365d3393555efc6548ce8eb1404c9df

I also just tested moving the `_prune_icons()` call just to make sure.

That said, `_prune_icons()` could perhaps be optimized. A low-hanging fruit might be to skip if the size of `_rendered_thumbnails` is below the unload threshold, before trying to scan through all the out of range thumbnails.

The more complicated optimization might be to look through rendered thumbnails, and check whether they are in range, instead of scanning through out of range thumbnails and seeing if they're loaded.

As for the other option you just described, I tried something where I tried to only call `_prune_icons` once, by keeping track of how many thumbnails were left to load. The behavior was not sufficiently consistent, though.

karlch:
Thanks for the hint! I like the idea of the low-hanging fruit(s). Another one would be to skip the call completely, if the settings are at default (i.e. `load_ahead` and `load_behind` are not `-1`). It might even be nice to have this whole logic only enabled, once the settings are changed :thinking:

But this is getting out-of-hand, and possibly too complicated alltogether. If you don't mind, I think I would merge this once all your commits are pushed (dict comprehension?) and we can continue thinking if we come up with something more efficient in another PR :blush:

buzzingwires:
By the time you sent your post, I honestly had already just about implemented the low-hanging fruit of checking if there were any thumbnails to be unloaded before scanning.

Your idea to check if load_ahead and load_behind are not -1 is already sort of implicitly done, in that retrieving the first or last rendered index will produce the first or last index of the list. So, when put into a range, it's empty and the loop should just quit as a result.

But I agree, this is starting to get into micro-optimization territory. I personally think the performance overhead *should* be fairly negligible like this.

Anyway, the dict comprehension has been pushed to, and we can merge the pull request here.

Though, there is a final bit of unrelated envelope pushing I feel compelled to mention: The way the pruning is done is from 'left to right'. When the unload threshold is exceeded there's a strong bias to pruning the thumbnails at the beginning of the list, before considering ones at the end of the list. It might be fairer to alternate between thumbnails at the beginning and end of the list, so either end is pruned roughly equally.

karlch:
Sorry for the delay :frowning_face: 

I finally got around to test the option of "only enabling pruning in case unload threshold > 0" in 1ecc08809f6478fd00e477612a25e9d0bb4e30f3. Instead of having the check in each call, we only connect the `_prune_icons` part once the `unload_threshold` exceeds `-1`. Like this, the whole logic is never touched in the default case. What do you think of this approach in general? Now, one could take this a few steps further, but I am not sure it is needed. Anyway, for completeness:
- Also check if either `load_ahead` or `load_behind` are `> -1`.
- Disconnect, once set manually back to default.

I see were you are coming from with the bias pruning thumbnails from the beginning. So instead of pruning ahead and behind, one would need to prune something along (there is probably a better way):
```python
    def _prune_icons(self) -> None:
        """Prune indexes before and after the current index."""
        if not self._prunable_thumbnails_exist():
            return
        idx_behind = range(0, self._first_rendered_index())
        idx_ahead = reversed(range(self._last_rendered_index() + 1, len(self._paths)))
        for idx in flatten(itertools.zip_longest(idx_behind, idx_ahead)):
            if idx is not None:  # None from zip_longest
                self._prune_index(idx)
```
With this general approach, one could probably also check `_prunable_thumbnails_exist` within the loop here, and return completely, instead of skipping each call to `_prune_index`?

UPDATE II: Not sure I got this, they would quit immediately as the first one has already cleared a sufficient amount, and thus `rendered + remaining < unload_threshold`?

buzzingwires:
I'm sorry if I pulled this PR out longer than intended. Again, we could totally just merge. We might be focusing too much on code that is already acceptably performant.

Checking `load_ahead` or `load_behind` seems like an even more extreme case of this, since again, the loops should not actually run if they're negative, though the functions will still be called.

I did look at the code you proposed. It looks sensible (and more consistent, since I have no idea why I checked `api.settings.thumbnail.unload_threshold.value == -1`) and like everything is still called in the right order(as in `_prune_icons` after `_on_thumbnail_created`). There are just a few things I thought to change.

Here, there's record keeping done to both disconnect when necessary, and to avoid multiple consecutive connects (which can produce extra calls to `_prune_icons` when `thumbnail.unload_threshold` is changed.).

```Python
    def _maybe_connect_pruning(self, unload_threshold: int):
        """Only connect pruning in case unloading should be performed."""
        if unload_threshold > -1:
            if not self._pruning_connected:
                self._manager.created.connect(self._prune_icons)
                self._pruning_connected = True
        elif self._pruning_connected:
            self._manager.created.disconnect(self._prune_icons)
            self._pruning_connected = False

```

There's also another `And I run set thumbnail.unload_threshold 0` in the `Display two thumbnails ahead and all behind` test scenario.

For fixing the bias, your code looks about correct to my initial observation and like it more or less produces the result I had in mind. I'd just want to test it. I probably don't have anything much better than that.

Finally, as for UPDATE II, honestly, I don't entirely remember what I meant. It's something I'd probably have to write a prototype for to properly conceptualize (and test) again, but I think I meant something that followed UPDATE I, if the length of `_rendered_paths` is less than the number of indexes to be scanned, by the pruning process, the loop could also quit early.

Yutsuten:
I have a photos folder with 1500+ pictures, and without this change my vimiv would always freeze pretty bad. Even after a few minutes it wouldn't respond, so I `kill` it.

I tried merging this change with v0.9.0. And used this configuration:

```
[THUMBNAIL]
save = False
size = 256
load_behind = 30
load_ahead = 30
unload_threshold = -1
```

This is what makes sense to me. No freeze. I'm able to navigate the thumbnails with no issue. Excellent work done here, good job!! :+1: 

I'm going to use this for now on! :smile:

PR #669, opened 2023-07-18
Fix all XML files detected as SVG

jcjgraf:
All XML files were detected as SVGs. This bug has been in vimiv ever since SVGs are supported (441b2d). Stupidly, I just copied to old SVG check function in #650 without thinking 😬

A xml file (i.e. a file starting with `<?xml`) is a svg, if and only if it is followed by a `<svg` tag[1].

[1] Do not quote me on this, standards are weird and the signature differentiates between different SVG standard versions. It is likely that we still have false negative rate, but we better only extend the detected signatures once we encounter such samples out in the wild...

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -6,6 +6,11 @@ All notable changes to vimiv are documented in this file.
 v0.10.0 (unreleased)
 --------------------
 
+Fixed:
+^^^^^^
+
+* The SVG image header check to no longer return true for all XML files.
+
 
 v0.9.0 (2023-07-15)
 -------------------
--- a/vimiv/utils/imageheader.py
+++ b/vimiv/utils/imageheader.py
@@ -40,9 +40,16 @@ Extended QT Support:
 | MNG    | TODO: ?                   | yes                                             |
 | TGA    | tga                       | yes (only version 2, version 1 is undetectable) |
 | TIFF   | tif, tiff                 | yes                                             |
-| WBMP   | wbmp                      | no (is undetectable)                      |
+| WBMP   | wbmp                      | no (is undetectable)                            |
 | WEBP   | webp                      | yes                                             |
 | ---    | ---                       | ---                                             |
+
+The docstrings of the verification functions show the bytes that are used to detect the
+file. Plain numbers are in base16. Two question marks (??) represent a variable byte.
+
+The bytes are also listed when interpreted as alphanumerical characters. Two dots (..)
+represent a non-alphanumerical character. Two question marks (??) represent a variable
+byte.
 
 A great list of magic bytes is provided here:
 https://en.wikipedia.org/wiki/List_of_file_signatures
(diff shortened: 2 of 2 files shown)


vimiv/utils/imageheader.py
```diff
@@ -171,26 +184,44 @@ def _test_gif(h: bytes, _f: BinaryIO) -> bool:
...
+    if h[2:5] == b"\x73\x76\x67":
+        return True
+
+    # Check for svg start tag
+    import re
+
+    f.seek(0, 0)
+    h = f.read()
+    pattern = b"<svg"
+    return re.search(pattern, h) is not None
```
karlch, line 224: I suppose we need the full-blown re search, as there can be all sorts of stuff in front (comments, newlines, spacing, <?xml, ...)?

docs/changelog.rst
```diff
@@ -6,6 +6,11 @@ All notable changes to vimiv are documented in this file.
 v0.10.0 (unreleased)
 --------------------
 
+Fixed:
+^^^^^^
+
+* The image header check for SVG was faulty and returned true for all XML files.
```
karlch, line 12: This sounds a bit harsh, but feel free to keep :blush: Would prefer something along:
Improved the SVG image header check to no longer return true for all XML files.

karlch (commented):
Thanks for finding this funky behaviour! Left a small question, but am basically happy to merge as-is.

vimiv/utils/imageheader.py
jcjgraf: A full search opposed to a strict matching (i.e. something like `<?xml>\n<svg>`) you mean? I presume that this does not not work in all cases. I have also seen SVGs that have multiple `<svg` tag. Therefor, I thought that there may be also some other xml tags in front of it, or that the `<svg` tag is at a higher depth. But I am by no means a specialist...

docs/changelog.rst
jcjgraf: I have improved the wording

vimiv/utils/imageheader.py
jcjgraf: Once this is resolved, the PR is mergeable.

karlch: Finding them is definitely a mess, I guess using `re` is the way to go, a random file I had around e.g. started with:
```
<?xml version="1.0" encoding="utf-8" standalone="no"?>
<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN"
  "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
<!-- Created with matplotlib (https://matplotlib.org/) -->
<svg height="255pt" version="1.1" viewBox="0 0 660 255" width="660pt" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
```

PR #674, opened 2023-07-31
Fix wrong statement that `imageformats` supports jp2

jcjgraf:
This should have been changed as part of #650.
This plugin is no longer needed for jp2 support, but jp2 is natively supported, given that the qt-imageformats module is installed.


karlch:
Thanks! :blush:

Issue #682, opened 2023-08-12
`<escape>` and `del` bindings don't work?


xfzv:
I'm trying to bind the `<escape>` and `del` keys in `keys.conf`, but they don't work for any command.

```console
❯ vimiv --version
vimiv v0.9.0

Python: 3.11.3
Qt: 5.15.8
PyQt: 5.15.9

Svg Support: True


Paths:
cache: /home/xfzv/.cache/vimiv
config: /home/xfzv/.config/vimiv
data: /home/xfzv/.local/share/vimiv
```

I installed `vimiv-qt` via [`vimviv-qt` (AUR)](https://aur.archlinux.org/packages/vimiv-qt).

- `~/.config/vimiv/keys.conf`

```console
[GLOBAL]
q : quit            # default - works
<escape> : quit     # doesn't work

[IMAGE]
dD : delete --ask %%    # works
del : delete --ask %%   # doesn't work

<end> : goto -1     # default - works
del : goto -1       # doesn't work

[COMMAND]
<escape> : leave-commandline    # default - works
```

I found <https://github.com/karlch/vimiv-qt/blob/036e41561456387d8b11895ca9a45428563007aa/vimiv/gui/eventhandler.py#L131-L137> regarding the `<escape>` key, maybe that's a limitation?

I couldn't find anything about `del` key though, I assume it should work. Both keys seem to be detected fine with `--debug gui.eventhandler`:

```console
❯ vimiv --debug gui.eventhandler
[..]
[12:10:10] DEBUG    <gui.eventhandler>   KeyPressEvent: handling <escape> key specially
[12:10:13] DEBUG    <gui.eventhandler>   EventHandlerMixin: handling del for mode image
[12:10:13] DEBUG    <gui.eventhandler>   EventHandlerMixin: no matches for event
```

karlch:
Thanks for catching this!

`<escape>` is currently a limitation, you are right, as it is hard-coded exactly like you found.

`del` doesn't work because vimiv interprets this as `d` `e` `l` in sequence. It should be added to the  `special_keys` dict in `gui.eventhandler` like so:
```python
        Qt.Key_Delete: "<delete>",
```
and then one can bind `<delete>` instead and it should work just fine :blush: 

Feel free to open a PR if you like, as you did all the debugging, otherwise I'll try to commit sometime later. I'll be offline for most of the week-end unfortunately.

karlch:
Perfect, thanks for the PR!

[...]

PR #691, opened 2023-08-15
Add :cancel command bound to <escape>

karlch:
The cancel command emits a generic api.signals.cancel signal. This has two advantages:

1) <escape> is no longer handled specially via hard-code.
2) The eventhandler does not have to explicitly know "what to cancel".
   In addition, plugins can also connect to api.signals.cancel in case
   they need some sort of "resetting".

Related to #682 

@xfzv this should allow binding `<escape>` as usual, do make sure to bind `:cancel` to something else though. Honestly, I don't quite see why one would want to do this, e.g. what will you bind  to "abort" keysequences such as `g?`? But the overall flexibility of the signal was on my mind in any case for some time.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -16,6 +16,9 @@ Added:
   * ``6`` or ``PyQt6``: Use PyQt6.
   * ``PySide6``: Use PySide6 (Qt for Python). This is highly experimental and should be
     used with care.
+* The generic ``:cancel`` command bound to ``<escape>`` together with the
+  ``api.signals.cancel`` signal which can be used by plugins in case they need some sort
+  of resetting.
 
 Fixed:
 ^^^^^^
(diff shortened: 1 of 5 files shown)


karlch:
Thanks for clarifying. That is kind of what I expected, also as it is pretty much implied by the `closeFullScreenOrExit` naming used by `qimgv`.

Pop-up, command line exit and so forth are not an issue, but everything that is in the same mode (global, i.e. image, thumbnail and library) is. And we do already have two things to cancel (any existing search results, and the key sequences), instead of one (fullscreen). Having those somehow communicating back does seem pretty intermingled, and removes any benefits of the decoupling just performed with the signal :smile: 

I would still consider merging this for consistency, and the signal-related decoupling in itself. But hope you are fine with not being able to have the `cancelOrQuit`-like functionality in vimiv, especially as you mention that `q` and `:q` are the "vim-like" standard :blush:

karlch:
Closing as we finally discussed having the command is not useful, and I merged the signal in #790 to allow plugins to connect to a "generic cancel" which is now always emitted by escape.

PR #692, opened 2023-08-15
Read binary image from stdin

### LINKED ISSUE #459: Add option to read image from standard in
It would be great if there was a command line flag like `-` to read a raw image content from stdin and display it.

Use cases:
Making images with Imagemagick:
```
convert -pointsize 100 label:"text" bmp:- | vimiv -
```
Rendering pic pictures:
```
pic file.pic | pic2graph | vimiv -
```
Now the user wouldn't need to create a temporary file each time she wants to recompile the image

Similar applications like `zathura` have this option

Here's a very simple diff of commit 4a1a0d1 that does this by creating a file in the vimiv cache directory: http://ix.io/2NMG/, but I think a better approach should be made

### LINKED ISSUE #678: suggestion: open a URL
it can be useful to be able to open URLs directly, i.e.:

`vimiv https://i.imgur.com/somefile.png`

karlch:
Finally adds the option to read a binary image from stdin using `vimiv -`.

fixes #459 
fixes #678 as adding an explicit option  to handle urls in vimiv seems over-the-top, but this allows the workaround of using `curl` to pipe to `vimiv -`.


karlch:
Thanks for your thoughts!

I fully agree with the mutually exclusive group, good catch, changed and pushed.

However, I am not convinced mingling the inputs is the way to go. While there really is no harm, I like the clarity of having one source of path(s) and one source only. Can you think of any use-case of actually passing both a flag and positional arguments or is this just an "add because it is in principle possible"? Unfortunately, it doesn't seem trivial to make positional arguments mutually exclusive with anything else. So, in case we want more clarity to the user, we would either have to clarify the wording / documentation or even raise a warning in case `-i` or `-` is passed, and there are positional arguments. Any thoughts?

jcjgraf:
> Can you think of any use-case of actually passing both a flag and positional arguments or is this just an "add because it is in principle possible"?

It is more of a "add because I did not figure out a way to have a mutually exclusive group with positional arguments" 😆 And no, I do not have a use-case for this either.
Maybe it is best to undo the changes of the last commit and instead add a custom mutually exclusive check, as you suggested yourself. E.g. like:

```diff
diff --git a/vimiv/startup.py b/vimiv/startup.py
index 886c4d04..96ec93bc 100644
--- a/vimiv/startup.py
+++ b/vimiv/startup.py
@@ -63,7 +63,12 @@ def setup_pre_app(argv: List[str]) -> argparse.Namespace:
     Args:
         argv: sys.argv[1:] from the executable or argv passed by test suite.
     """
-    args = parser.get_argparser().parse_args(argv)
+    argparser = parser.get_argparser()
+    args = argparser.parse_args(argv)
+    if sum([args.paths != [], args.stdinput, args.binary_stdinput]) > 1:
+        argparser.error(
+            "positional argument [PATH...], argument -i/--input, and argument - are mutually exclusive."
+        )
     if args.version:
         import vimiv.version
```

karlch:
Thanks! I agree this is probably the best way forward, will incorporate once I get to it and merge then :blush:

PR #695, opened 2023-08-18
Fix empty lines in tag file  

jcjgraf:
It is possible to insert empty lines into the tag files. Once the tagged images of that tag are loaded again, an empty string is inserted into the marked list. This leads to issues (e.g. `:open :m` fails).

## How to insert empty lines into the tag file
Simply run `:tag-write test` multiple times without adding marking a new image in-between.

## This PR fixes the issue at several levels
- Prevent insertion of empty lines into the tag file: 3561980c3a61f2e615c6009bfd2b65cf4400e7bb
- Prevent empty line getting read tags: 9b657c44979253e24147d8f94afa323a338eae9c [1]

[1] 96abdba02681d0f6716e8b4b00f38359555e4efa would also have fixed the second point. But I added that commit rather to get rid of repeated code. In addition, it ensures that images loaded from the tag are actually also added to the watcher, which they are not without this commit.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -30,6 +30,8 @@ Fixed:
 
 * The SVG image header check to no longer return true for all XML files.
 * Binding the ``<delete>`` key as special key. Thanks `@xfzv`_!
+* Consecutive `:tag-write` would insert empty lines into the tag file, which would on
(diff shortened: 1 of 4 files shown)


karlch:
Thank you, this one looks good to me as-is! :blush:

PR #705, opened 2023-08-23
Add support for long keybindings

jcjgraf:
This PR adds support for keybindings of arbitrary-length key sequences. Longer key sequences were actually *almost* supported, but due two bugs (probably you just never really thought about them) prevented their (proper) working.

## How to test
Adds support for bindings like `:bind abc log info success`. Without this PR, such a binding is not work

## Changes
Two things prevented the working of such a binding and required adoption:

#### Problem
`gui.eventhandler.EventHandlerMixin._process_event()` is called for each keystroke. That function calls `self.partial_handler.keys.get_keys()`, which in turn calls `self._keys.clear()`. On a partial match, `self.partial_handler.keys.add_keys(*sequence)` is run.

This means that for each new keypress the keys from the keystore are retrieved and the store is wiped. On a partial match, `sequence`, which is only the last pressed key, is added to the store again. This means, that only one key is stored in the keystore at the time.

#### Fix
I fixed that issue by removing `self._keys.clear()` in `TempKeyStorage.keys.get_keys()`. I really do not see why this call should be required. If however it is, the issue could also be fixed by adding not only `sequence`, but actually the *full sequence* (all previously pressed keys) to the keystore.

### Keyhint shows wrong "already typed keys" (aka. prefix)
The keyhint always displays the most recently pressed key as the prefix and not the whole sequence of pressed keys.

#### Problem
In `gui.eventhandler.EventHandlerMixin._process_event()` on a partial match, the signal `partial_matches` is emitted. As argument `name` is passed. However, `name` is based solely on `sequence`, which is the currently pressed key` and not the whole sequence of pressed keys.

#### Solution
Pass all keys retrieved from the keystore as the first argument to the signal.


karlch:
It's been a while, but I finally did some more testing and playing around, and to me this seems superior to the current behaviour in every way. So if you give me a quick thumbs up, happy to merge :blush:

PR #749, opened 2023-12-05
Detect SVGs that do not start with a header

### LINKED ISSUE #748: Some SVG files are not detected
Thanks a lot for this amazing application. It's by far my favorite image viewer.

I noticed that Vimiv refuses to open a lot of SVG files. A sample file that isn't recognized on my system: https://upload.wikimedia.org/wikipedia/commons/4/4f/SVG_Logo.svg   

Adding `<?xml version="1.0" encoding="UTF-8"?>` to the beginning of the file seems to fix this issue. Is there maybe a way to force Vimiv to open all `*.svg` files?

```
$ vimiv --version
vimiv v0.9.0

Python: 3.10.12
Qt: 5.15.3
PyQt: 5.15.6

Svg Support: True
```

jcjgraf:
Indeed, there are (actually many) SVGs that do not start with a `<?` header but simply `<svg`. I have added detection for them.Though, I still did not manage to find documentation on all possible header of SVG files. What when the `<svg` line is preceded by a comment line? `file` detects such SVGs as HTML.

@infinitewhileloop Thanks for reporting and the submission of the sample.

fixes #748

### DIFF
--- a/vimiv/utils/imageheader.py
+++ b/vimiv/utils/imageheader.py
@@ -208,9 +208,14 @@ def _test_svg(h: bytes, f: BinaryIO) -> bool:
      ->  <  ?  x  m  l
     --> 3C 3F 73 76 67 (with stripped xml header)
      ->  <  ?  s  v  g
+    --> 3C 73 76 67 (inline without header)
+     ->  <  s  v  g
 
     Native QT support.
     """
+    if h[:4] == b"\x3C\x73\x76\x67":
+        return True
+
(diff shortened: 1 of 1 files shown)


karlch:
Awesome, thanks a lot for looking into this and fixing the issue!

The lack of a standard does seem annoying, and I guess there really is nothing we can do about this if even `file` does not identify some "valid" svg files correctly.
As long as we get a good amount I guess we are fine :blush:

PR #755, opened 2023-12-15
Add more supported formats in vimiv.desktop

### LINKED ISSUE #754: [feature request] add missing mime types to the desktop entry file
In the `vimiv.desktop`, there are only a few image formats:
```
MimeType=image/bmp;image/gif;image/jpeg;image/jp2;image/jpeg2000;image/jpx;image/png;image/svg;image/tiff;
```
Some supported formats, e.g., `image/webp`, are missing from the list. As a result, vimiv won't be shown in the right-click menu or the recommended "open-with" list in GUI file managers. Adding those to the desktop entry file will be great.

xlucn:
Closes #754 (Oops, I forgot to add this in the commit message)

Supported formats according to [this](https://github.com/karlch/vimiv-qt/blob/master/vimiv/utils/imageheader.py#L11C1-L25C85) (referring to [the comment](https://github.com/karlch/vimiv-qt/issues/754#issuecomment-1856502245)) are added in the desktop entry file.

I have included the formats supported by the `qtimageformats` module. If that's an issue, I think it can be mitigated by showing a in-program warning to install optional dependencies. However, not including them will not solve the issue and can't be mitigated later.

### DIFF
--- a/misc/vimiv.desktop
+++ b/misc/vimiv.desktop
@@ -7,4 +7,4 @@ Icon=vimiv
 Terminal=false
 Exec=vimiv %F
 Categories=Graphics;
-MimeType=image/bmp;image/gif;image/jpeg;image/jp2;image/jpeg2000;image/jpx;image/png;image/svg;image/tiff;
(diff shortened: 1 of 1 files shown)


karlch:
Awesome, thank you, I fully agree with your logic in having the additional formats in the desktop file - as it is static and we can't really extend it in the code.

